PR #52, opened 2014-09-25
[parseq-23] add seq() convenience method for list of Tasks plus a single Task

innerverse:
[parseq-23] add seq() convenience method for list of Tasks plus a single Task


jodzga:
API in version 2.x has changed so this issue does not apply to current state of ParSeq.

PR #53, opened 2014-10-09
Add Tasks.value for making tasks that trivially return a value.

ts4z:
ValueTask appears to be in the test code. I need something like this, and right now I'm having to use Tasks.callable("...", new Callable() { Whatever call() { return value; }); all over the place (actually, a wrapper around that).

I'd like to suggest this change to move it somewhere accessible.  I can provide some examples within LinkedIn source where I think this is a useful thing to do.


ts4z:
Task.value in parseq 2 is exactly what I wanted.

PR #54, opened 2014-11-11
Tasks.<T>par(emptyIterable<? extends Task<? extends T>>) should not throw IllegalArgumentException

jdowell:
This seemed like an easy feature for a first commit, but please let me know if you'd like to see other changes/testing here.


jodzga:
Solved by #113.

PR #62, opened 2015-09-30
Batching implementation prototype

jodzga:
Design is not finalized yet. This pull request is to collaborate and discuss.


contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+
+      Long planId = ctx.getPlanId();
+      List<InternalBatchEntry<K, Object>> list = batches.get(planId);
+      if (list == null) {
+        list = new ArrayList<>();
+        List<InternalBatchEntry<K, Object>> existing = batches.putIfAbsent(planId, list);
+        if (existing != null) {
+          list = existing;
+        }
+      }
```
stevenheidel: I think you can avoid two calls to the ConcurrentHashMap here:

```java
Long planId = ctx.getPlanId();
List<InternalBatchEntry<K, Object>> newList = new ArrayList<>();
List<InternalBatchEntry<K, Object>> list = batches.putIfAbsent(planId, newList);
if (list == null) list = newList;
```

Took me a while to figure out that you were doing the second `existing != null` check because of the potential race condition with something getting added to the batch between `.get` and `.putIfAbsent`.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingSupport.java
```diff
@@ -0,0 +1,27 @@
...
+  public void registerStrategy(BatchingStrategy<?, ?> strategy) {
+    _strategies.add(strategy);
+  }
+
+  @Override
+  public void onPlanActivated(final PlanContext plaContext) {
+  }
+
+  @Override
+  public void onPlanDeactivated(final PlanContext plaContext) {
```
stevenheidel: planContext

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
```diff
@@ -0,0 +1,70 @@
...
+  public static class ExampleBatchingStrategy extends BatchingStrategy<Long, Long> {
+
+    @Override
+    public Long groupForKey(Long key) {
+      //lets say all tasks can be grouped into one batch
+      return 0L;
+    }
+
+    @Override
+    public Task<?> taskForBatch(final Long group, final List<? extends BatchEntry<Long, Object>> batch) {
```
stevenheidel: Is there a way to not lose the type information here (Object)? Seems like in the case of rest.li you'd end up casting these to Get requests or BatchGet requests  all the time.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+      new ContextImpl(planContext, task).runTask();
+    } catch (Throwable t) {
+      failAll(batch, t);
+    }
+  }
+
+  public abstract G groupForKey(K key);
+
+  public abstract Task<?> taskForBatch(G group, List<? extends BatchEntry<K, Object>> batch);
```
stevenheidel: Why does this return `Task<?>` instead of `Task<Map<K, Object>>`?

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @author Jaroslaw Odzga (jodzga@linkedin.com)
+ */
+public class BatchingClientExample extends AbstractExample {
+
+  public static class ExampleBatchingStrategy extends BatchingStrategy<Long, Long> {
+
+    @Override
+    public Long groupForKey(Long key) {
```
stevenheidel: I'm not convinced this is the best way of doing grouping. It seems to me like you'd have two very common cases:

1) All the keys can be batched together. For instance, in the case of fetching profiles we would always have the profile id. We would write the batch get method to take a list of profile ids and fetch a list of profiles. In this case it's just extra work to override the `groupForKey` method.

2) The keys have some additional parameter added to them. For instance, you want to get a profile given a profile id and a viewer id. You can only batch get profile ids together if they have the same viewer id. Therefore, your key would be a tuple (viewer id, profile id). Given a list of tuples like:
(A, 1)
(A, 2)
(B, 1)
you'd group them into:
(A, List(1,2))
(B, List(1))
In this case you'd want the group type G to be a single object and the taskForBatch method to take G and a List of profile ids, NOT a list of tuples (because all the first elements in the tuple would be the same).

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
```diff
@@ -0,0 +1,70 @@
...
+  Task<String> batchableTask(final Long id) {
+    return batchingStrategy.batchable("fetch id: " + id, () -> {
+      return new BatchEntry<Long, String>(id, Promises.settable());
+    });
+  }
+
+  @Override
+  protected void customizeEngine(com.linkedin.parseq.EngineBuilder engineBuilder) {
+    BatchingSupport batchingSupport = new BatchingSupport();
+    batchingSupport.registerStrategy(batchingStrategy);
```
stevenheidel: So you'll be able to register multiple different things that can be batched and we'll figure out the correct one based on the plan context? Have I understood this correctly?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: Yeah, that was actually my first implementation. That form however makes one additional allocation and I think (haven't tested it yet) JVM would not be able to optimize it away because it is being passed in putIfAbsent() method. I decided to trade one extra ArrayList allocation on every call for one extra ConcurrentHashMap lookup on first call for the plan.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingSupport.java
jodzga: Thanks :-)

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
jodzga: This is tricky. Not really sure how this will play out. Will need to prototype it for some types of tasks e.g. rest.li.
The reason for downcasting everything to Object is that you may want to be able to batch together tasks that don't have any type relationships. For example, if we used multiplexed reat.li requests we might be able to batch together not only GET operations but really arbitrary operations.
I think it is necessary to prototype few implementations first to see how this API would play out.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: You guessed my intentions correctly - sorry for not documenting this clearly: the idea is that client would be responsible to completing promisses.
One aspect of this design needs definitely work - error handling and marking batch request as a completed. It would be good to have some contract around this e.g. entire batch task is completed once all promises are completed. This needs to be iterated over and improved.

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
jodzga: I agree. Perhaps we could provide BatchingStrategy that would be most general and would not do any grouping and a subclass GroupingBatchingStrategy that would be more convenient for grouping (it would contain changes you are mentioning).

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/batching/BatchingClientExample.java
jodzga: You'll be able to register multiple things that can be batched e.g. rest.li parseq client would register one strategy, voldemort parseq client would register one as well, etc. We would not have to figure out anything because batching implementation would be part of task execution - parseq would be completely oblivious to batching.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
stevenheidel: I think you're right in that a good design will emerge based on actually trying this out in practice. I think we'll get more clarity on this question and the downcasting to Object question after our meeting tomorrow re: rest.li multiplexing.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
```diff
@@ -0,0 +1,148 @@
...
+import com.linkedin.parseq.promise.Promises;
+import com.linkedin.parseq.promise.SettablePromise;
+import com.linkedin.parseq.trace.ShallowTraceBuilder;
+
+public class BatchImpl<K, T> implements Batch<K, T> {
+
+  private final Map<K, BatchEntry<T>> _map;
+
+  private BatchImpl(Map<K, BatchEntry<T>> map) {
+    _map = map;
```
zhenkai: should copy

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Override
+  public void done(K key, T value) throws PromiseResolvedException {
+    _map.get(key).getPromise().done(value);
```
zhenkai: check for null?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    _map.forEach((key, entry) -> consumer.accept(key, entry.getPromise()));
+  }
+
+  @Override
+  public String toString() {
+    return "BatchImpl [entries=" + _map + "]";
+  }
+
+  @Override
+  public boolean failAllRemaining(Throwable error) {
```
zhenkai: what does this do? fail all unsatisfied promised? why is it failing all tasks (including those already set)?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+import com.linkedin.parseq.trace.Relationship;
+import com.linkedin.parseq.trace.ShallowTraceBuilder;
+import com.linkedin.parseq.trace.TraceBuilder;
+
+public abstract class BatchingStrategy<G extends Group, K, T> {
+
+  private final ConcurrentHashMap<Long, BatchBuilder<K, T>> _batches =
+      new ConcurrentHashMap<>();
+
+  public Task<T> batchable(final String desc, final K key) {
```
zhenkai: does this mean "createTask"?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  public abstract void executeBatch(G group, Batch<K, T> batch);
+
+  public abstract void executeSingleton(G group, K key, BatchEntry<T> entry);
+
+  public abstract G classify(K entry);
+
+  public Map<G, Batch<K, T>> split(Batch<K, T> batch) {
```
zhenkai: can be private?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+        return Batch::<K, T>builder;
+      }
+
+      @Override
+      public BiConsumer<BatchBuilder<K, T>, Entry<K, BatchEntry<T>>> accumulator() {
+        return (builder, entry) -> builder.add(entry.getKey(), entry.getValue());
+      }
+
+      private BatchBuilder<K, T> combine(BatchBuilder<K, T> larger, BatchBuilder<K, T> smaller) {
+        return null;
```
zhenkai: ?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
```diff
@@ -0,0 +1,251 @@
...
+import com.linkedin.restli.internal.common.ProtocolVersionUtil;
+import com.linkedin.restli.internal.common.ResponseUtils;
+
+public class GetRequestGroup implements RequestGroup {
+
+  private final String _baseUriTemplate;
+  private final Map<String, String> _headers;
+  private final Map<String, Object> _queryParams;
+
+  public GetRequestGroup(Request request) {
```
zhenkai: Request is parameterized below?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
```diff
@@ -0,0 +1,251 @@
...
+  private static Map<String, Object> getQueryParamsForBatchingKey(Request<?> request)
+  {
+    final Map<String, Object> params = new HashMap<String, Object>(request.getQueryParamsObjects());
+    params.remove(RestConstants.QUERY_BATCH_IDS_PARAM);
+    params.remove(RestConstants.FIELDS_PARAM);
+    return params;
+  }
+
+
+  <RT extends RecordTemplate> List<BatchGetRequest<RT>> toTypedList(List l) {
```
zhenkai: what is this...

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
```diff
@@ -0,0 +1,251 @@
...
+    //TODO this needs better API from rest.li team
+    Class<?> keyClass = batch.entires().iterator().next().getKey().getRequest().getResourceSpec().getKeyClass();
+    if (CompoundKey.class.isAssignableFrom(keyClass) || keyClass == ComplexResourceKey.class) {
+      executeKVBatch(restClient, batch);
+    } else {
+      executeNonKVBatch(restClient, batch);
+    }
+  }
+
+  static GetRequest toGetRequest(Request request) {
```
zhenkai: unsafeToGetRequest

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,158 @@
...
+          //all we care is that all remaining promises have been failed
+          batch.failAllRemaining(t);
+        }
+      }
+    }
+  }
+
+  public abstract void executeBatch(G group, Batch<K, T> batch);
+
+  public abstract void executeSingleton(G group, K key, BatchEntry<T> entry);
```
zhenkai: I'm a little concerned about this API. It seems BatchEntry is more like a class for parseq internal use, but here we expose it to user code. This leaks too much (e.g. addShallowTraceBuilder, getShallowTraceBuilder). 

Can we use use SettablePromise here?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
```diff
@@ -0,0 +1,36 @@
...
+
+  void fail(K key, Throwable error) throws PromiseResolvedException;
+
+  void failAll(Throwable error) throws PromiseResolvedException;
+
+  boolean failAllRemaining(Throwable error);
+
+  Collection<BatchEntry<T>> values();
+
+  Set<Map.Entry<K, BatchEntry<T>>> entires();
```
zhenkai: do we have to expose those?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -0,0 +1,214 @@
...
+import com.linkedin.parseq.batching.BatchingStrategy;
+import com.linkedin.parseq.promise.Promise;
+import com.linkedin.parseq.promise.Promises;
+import com.linkedin.parseq.promise.SettablePromise;
+import com.linkedin.r2.message.RequestContext;
+import com.linkedin.restli.common.OperationNameGenerator;
+
+
+public class ParSeqRestClient
+    extends BatchingStrategy<RequestGroup, ParSeqRestClient.RestRequestBatchKey, Response<Object>> {
```
zhenkai: it's a bit weird for a Client to extend Strategy... maybe we can make the strategy a member of the client class?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/RequestGroup.java
```diff
@@ -0,0 +1,24 @@
...
+  public static RequestGroup fromRequest(final Request<?> request) {
+    switch (request.getMethod()) {
+      case GET:
+        return new GetRequestGroup(request);
+      default:
+        return new SingletonRequestGroup(ParSeqRestClient.generateTaskName(request));
+    }
+  }
+
+  abstract <RT extends RecordTemplate> void executeBatch(RestClient restClient, Batch<RestRequestBatchKey, Response<Object>> batch);
```
zhenkai: abstract? this is an interface...

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/RequestGroup.java
```diff
@@ -0,0 +1,24 @@
...
+
+  abstract <RT extends RecordTemplate> void executeSingleton(RestClient restClient, RestRequestBatchKey key, BatchEntry<Response<Object>> entry);
```
zhenkai: ditto

src/com/linkedin/parseq/Engine.java
```diff
@@ -204,6 +205,14 @@ public boolean awaitTermination(final int time, final TimeUnit unit) throws Inte
     return _terminated.await(time, unit);
   }
 
+  public boolean addPlanActivityListener(final PlanActivityListener planActivityListener) {
+    return _planActivityListeners.add(planActivityListener);
+  }
+
```
zhenkai: do we have a use case for this, adding listeners after construction of engine? If not, we can remove these two methods.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
jodzga: Why? Copy is expensive.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
jodzga: do you mean check that _map.get() != null, what should be behavior when _map.get() == null?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
jodzga: API definitely needs documentation. It fails all unsatisfied promises and returns true if ann promises were failed. It returns false if there was at least one promise that has already been satisfied.
If promise is already resolved then trying to fail it will throw exception but it will not fail it.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: batchable() returns new task that may be batched by the strategy but this is what consumer gets.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: yup

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: probably should throw exception here

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
jodzga: Generifying still needs more work.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
jodzga: This is fun working with Java generics ;-)

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
jodzga: Can you explain what do you mean?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: Can you branch and show what you mean?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
jodzga: I think clients will need those functions.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
jodzga: Sounds like a good idea.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/RequestGroup.java
jodzga: yup

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/RequestGroup.java
jodzga: yup

src/com/linkedin/parseq/Engine.java
jodzga: The problem is that if we don't make this dynamic, then we impose the order in which component have to be constructed i.e. all listeners have to be constructed before engine. This is OK if we are the only ones adding listeners but this will prevent 3rd party listeners to be added to engine.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Override
+  public void fail(K key, Throwable error) throws PromiseResolvedException {
+    _map.get(key).getPromise().fail(error);
+  }
+
+  @Override
+  public void failAll(Throwable error) throws PromiseResolvedException {
+    PromiseResolvedException exception = null;
```
angxu: Why not use MultiException?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+  }
+
+  public static class BatchBuilder<K, T> {
+
+    private final Map<K, BatchEntry<T>> _map = new HashMap<>();
+
+    public BatchBuilder<K, T> add(K key, BatchEntry<T> entry) {
+      //deduplication
+      BatchEntry<T> duplicate = _map.get(key);
```
angxu: can be _map.put(key).

src/com/linkedin/parseq/internal/SerialExecutor.java
```diff
@@ -85,6 +97,10 @@ public void run() {
       try {
         runnable.run();
       } finally {
+        if (_pendingCount.get() == 1) {
+          _active = false;
+          _activityHandler.deactivated();
+        }
```
angxu: This is not thread-safe, right? Someone can increment the _pendingCount after this block and have you tryExecuteLoop() again.

src/com/linkedin/parseq/Engine.java
zhenkai: Yeah, that's a tradeoff... I think it would be easier to reason/debug if the listeners are provided at construction time.  But if we don't have this use case, probably we can add the dynamic part later when there is a request/need, since it's backward compatible change.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
zhenkai: yeah, my question is basically what if user calls done on a non-exist key so that _map.get(key) is null? shall we also provide a boolean contains(K key) method like the Map?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
zhenkai: I think we could avoid unnecessarily throwing PromiseResolvedException given that from the method name it is expected some tasks could be finished. Currently it would try to fail every task and thus result in throwing & catching PromiseResolvedException for those finished

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
zhenkai: basically, Is the cast from Request to GetRequest safe?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
zhenkai: will try :)

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
jodzga: Yes because we are in the GetRerquestGroup that handles only GET requests.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
jodzga: Probably should use MultiException here.

src/com/linkedin/parseq/internal/SerialExecutor.java
jodzga: This is expected and thus thread-safe. If situation you described happens there will be the following sequence called:
1) deactivated()
2) activated()
I'll put more comments to explain what is going on here.

angxu: I see. Never mind. I somehow thought _activityHandler.deactivated() was running asynchronously...

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
stevenheidel: `entries` not `entires`

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,245 @@
...
+      return result;
+    });
+  }
+
+  private Collection<Task<?>> taskForBatches(Collection<Entry<G, Batch<K, T>>> batches) {
+    return batches.stream().map(entry -> taskForBatch(entry.getKey(), entry.getValue())).collect(Collectors.toList());
+  }
+
+  void handleBatch(final PlanContext planContext) {
+    final BatchBuilder<K, T> batchBuilder = _batches.remove(planContext.getId());
```
stevenheidel: So the BatchBuilder is added to _batches if at least one batchable task is created. And then here it gets removed, all Tasks in the same plan will have the same plan context, so this is guaranteed to work. Do I have that right?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,245 @@
...
+  private final ConcurrentHashMap<Long, BatchBuilder<K, T>> _batches =
+      new ConcurrentHashMap<>();
+
+  /**
+   * This method returns Task that returns value for a single key in allowing this strategy to batch operations.
+   * @param desc description of the task
+   * @param key key
+   * @return Task that returns value for a single key in allowing this strategy to batch operations
+   */
+  public Task<T> batchable(final String desc, final K key) {
```
stevenheidel: Are you against a `batchable(final K key)` method that generates a simple desc for you? Like "Task for id $key" or something

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -0,0 +1,219 @@
...
+   * @return response promise
+   */
+  @Deprecated
+  public <T> Promise<Response<T>> sendRequest(final Request<T> request, final RequestContext requestContext) {
+    final SettablePromise<Response<T>> promise = Promises.settable();
+
+    // wrapper around the callback interface
+    // when the request finishes, the callback updates the promise with the corresponding
+    // result
+    _restClient.sendRequest(request, requestContext, new PromiseCallbackAdapter<T>(promise));
```
stevenheidel: Ah, here's the Promise API for restliClient :)

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchImpl.java
jodzga: This can be achieved by querying set returned by keys(). I try to keep API to the bare minimum. Do you expect that one would need contains(K key) often?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
jodzga: Thanks!

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: Yup.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: I can add this version of batchable(). It would be consistent with approach we used in Task.* API where we provide two versions of methods: one where you provide description and one where we generate it for you.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param key key that identifies a Promise to be completed
+   * @param value value to complete Promise with
+   * @throws PromiseResolvedException if Promise associated with given key has already been resolved
+   */
+  void done(K key, T value) throws PromiseResolvedException;
+
+
+  /**
+   * Fails a {@link Promise} associated with given key with
+   * a an error.
```
stevenheidel: an error

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param error error to fail Promise with
+   * @throws PromiseResolvedException if Promise associated with given key has already been resolved
+   */
+  void fail(K key, Throwable error) throws PromiseResolvedException;
+
+  /**
+   * Fails all promises belonging to this batch with given error.
+   * This method throws PromiseResolvedException if any of the promises
+   * have already been resolved.
+   * This method guarantees that each Promise in this batch is either completed or failed.
```
stevenheidel: This could be made more clear, all remaining promises will be marked as failed even if one of them throws a PromiseResolvedException?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+import com.linkedin.parseq.promise.PromiseListener;
+import com.linkedin.parseq.promise.Promises;
+import com.linkedin.parseq.promise.SettablePromise;
+import com.linkedin.parseq.trace.Relationship;
+import com.linkedin.parseq.trace.ShallowTraceBuilder;
+import com.linkedin.parseq.trace.TraceBuilder;
+
+/**
+ * {@code BatchingStrategy} helps building "batching clients" in ParSeq. "Client" means on object that given {@code K key}
+ * provides a task that returns {@code T value}. "Batching" means that it can group together keys to resolve values
```
stevenheidel: BatchingStrategy helps **build** "batching clients" in ParSeq. "Client" means **an** object that, given K key provides a task that returns T value.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+ *    {@code @Override}
+ *    public Integer classify(Long entry) {
+ *      return 0;
+ *    }
+ *  }
+ * </pre></blockquote>
+ *
+ * In above example there is an assumption that all keys can be grouped together. This is why method {@code classify()}
+ * trivially returns a constant {@code 0}. In practice {@code classify()} returns an equivalence class. All keys that
+ * returns equal equivalence class will constitute a batch.
```
stevenheidel: I think it would be clearer to say "Only keys that have the same group will be batched together"

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+      if (batch.size() > 0) {
+        try {
+          final Map<G, Batch<K, T>> batches = split(batch);
+          if (batches.size() > 0) {
+            taskForBatches(batches.entrySet()).forEach(task -> new ContextImpl(planContext, task).runTask());
+          }
+        } catch (Throwable t) {
+          //we don't care if some of promises have already been completed
+          //all we care is that all remaining promises have been failed
+          batch.failAllRemaining(t);
```
stevenheidel: Does this fail all the groups or just the one that failed?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * This method will be called for every {@code Batch} that contains at least two elements.
+   * Implementation of this method should make sure that all {@code SettablePromise} contained in the {@code Batch}
+   * will eventually be resolved - typically asynchronously.
```
stevenheidel: Make this stronger: what happens if you don't resolve all the promises? Is the behaviour undefined or does the Task run indefinitely?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+   * @param key key of the single element classified to the given group
+   * @param entry entry contains a {@code SettablePromise} that eventually needs to be resolved - typically asynchronously
+   */
+  public abstract void executeSingleton(G group, K key, BatchEntry<T> entry);
+
+  /**
+   * Classify the {@code K Key} and by doing so assign it to a {@code G group}.
+   * If two keys are classified by the same group then they will belong to the same {@code Batch}.
+   * For each batch either {@link #executeBatch(Object, Batch)} will be called if batch contains at least two elements
+   * or {@link #executeSingleton(Object, Object, BatchEntry)} will be called if batch contains only one element.
```
stevenheidel: (G, Batch) and (G, K, BatchEntry) instead of Object?

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -0,0 +1,282 @@
...
+  public abstract G classify(K key);
+
+  /**
+   * Overriding this method allows providing custom name for a batch. Name will appear in the
+   * ParSeq trace as a description of the task that executes the batch.
+   * @param batch batch to be described
+   * @param group group to be described
+   * @return name for the batch and group
+   */
+  public String getBatchName(Batch<K, T> batch, G group) {
```
stevenheidel: In other methods, G group is the first parameter, should probably be consistent

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingSupport.java
```diff
@@ -0,0 +1,41 @@
...
+import com.linkedin.parseq.internal.PlanContext;
+import com.linkedin.parseq.internal.PlanDeactivationListener;
+
+/**
+ * This class allows registering instances of {@link BatchingStrategy}.
+ * <p>
+ * Please note that BatchingSupport must be registered with an {@link EngineBuilder} e.g.
+ * <blockquote><pre>
+ *  BatchingSupport batchingSupport = new BatchingSupport();
+ *  engineBuilder.setPlanActivityListener(batchingSupport);
```
stevenheidel: PlanDeactivationListener

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
jodzga: thanks!

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: I think I'll leave just failAll and remove failAllRemaining because the difference between them is too small to justify existence of both of them.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: sounds better, thanks

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: All groups are failed. This will happen only if there is a serious bug e.g. "classify" throws.
Tasks for each batch are only scheduled for execution here, not yet executed.

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: rephrased

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: Thanks, this was autocompleted ;-)

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingSupport.java
jodzga: thanks

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
jodzga: Unfortunately javadoc will actually fail if I use generics here so I'll bring back Object here.

contrib/parseq-batching/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+```java
+MyBatchingStrategy myBatchingStrategy = new MyBatchingStrategy();
+_batchingSupport.registerStrategy(myBatchingStrategy);
+```
+
+BatchingStrategy
+================
+
+```BatchingStrategy``` allows building "batching clients" where "client" means on object that given ```K key``` provides a task that returns ```T value```. ```BatchingStrategy``` defines which keys can be grouped together into batches and how batches are executed.
```
stevenheidel: means **an** object

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.function.BiConsumer;
+
+import com.linkedin.parseq.batching.BatchImpl.BatchBuilder;
+import com.linkedin.parseq.batching.BatchImpl.BatchEntry;
+import com.linkedin.parseq.promise.Promise;
+import com.linkedin.parseq.promise.PromiseResolvedException;
+import com.linkedin.parseq.promise.SettablePromise;
+
+/**
+ * Batch represents a collection of keys related to each other in such a way that
```
stevenheidel: that **it** is more efficient

contrib/parseq-batching/README.md
jodzga: Thanks!

contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/Batch.java
jodzga: Thanks!

PR #64, opened 2015-10-28
Blocking engine

jodzga:
BlockingEngine is a wrapper around an Engine that allows throttling execution of tasks by blocking a thread that starts a new plan until Engine has a capacity to execute it. This type of throttling mechanism is useful in situations where synchronous, single-threaded task producer needs a blocking throttling e.g. in kafka consumer. If there are multiple single-threaded task producers, e.g. kafka consumers of different topics, they should all share one instance of BlockingEngine.

Example usage in kafka consumer:

 while (_iterator.hasNext()) {
   // get next kafka message
   Record record = _iterator.next();

   // create task that will process record
   Task<?> task = processRecord(record);

   // will block until engine has capacity to process the task
   _blockingEngine.run(task);
 }
 
BlockingEngine can be parameterized by specifying how many concurrent plans can be running at any given time. If tasks are doing only synchronous operations then good value for the max concurrency is number of cores. If tasks are involved in asynchronous operations such as IO then in order to maximize CPU utilization max concurrent plans should be higher than number of available cores because it is expected that some of the plans will be suspended waiting for IO operations to complete.
BlockingEngine also provides a FIFO queue in which tasks wait until Engine has capacity to execute them.

BlockingEngine is not intended to be used in reactive settings because it is blocking.

### DIFF
--- a/contrib/parseq-blocking-engine/src/main/java/com/linkedin/parseq/blocking/BlockingEngine.java
+++ b/contrib/parseq-blocking-engine/src/main/java/com/linkedin/parseq/blocking/BlockingEngine.java
@@ -1,0 +1,141 @@
+package com.linkedin.parseq.blocking;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.linkedin.parseq.Engine;
+import com.linkedin.parseq.Task;
+import com.linkedin.parseq.internal.ArgumentUtil;
+
+
+/**
+ * {@code BlockingEngine} is a wrapper around an {@code Engine} that allows throttling execution of tasks by
+ * blocking a thread that starts a new plan until Engine has a capacity to execute it.
+ * This type of throttling mechanism is useful in situations where synchronous,
+ * single-threaded task producer needs a blocking throttling e.g. in kafka consumer.
+ * If there are multiple single-threaded task producers, e.g. kafka consumers of different topics,
+ * they should all share one instance of {@code BlockingEngine}.
+ * <p>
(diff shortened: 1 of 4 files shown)


contrib/parseq-blocking-engine/src/main/java/com/linkedin/parseq/blocking/BlockingEngine.java
```diff
@@ -0,0 +1,138 @@
...
+ * This class is not intended to be used in reactive settings because it is blocking.
+ *
+ * @author Jaroslaw Odzga (jodzga@linkedin.com)
+ */
+public class BlockingEngine {
+
+  private final Engine _engine;
+  private final Semaphore _demand;
+  private final Queue<Task<?>> _queue = new ConcurrentLinkedQueue<>();
+  private final AtomicInteger _pending = new AtomicInteger(0);
```
zhenkai: what's the advantage of this over a bounded BlockingQueue?

jodzga: How would you use bounded blocking queue in this context?
Thread that completes plan (might be a parseq thread) needs to get next task from the queue for execution and we don't want to block it on reading from queue.

zhenkai: When submitting the task, use blocking api put to block user thread if needed, so user code don't submit too fast; when retrieving task, use non blocking api poll so the engine threads don't block. Would that work?

jodzga: Are you suggesting simply replacing ConcurrentLinkedQueue with BlockingQueue or getting rid of AtomicInteger and semaphore as well?
Maybe you could provide an alternative pull request? You could just check out this branch, modify it and create pull request from it? This should be not much work...

PR #65, opened 2015-11-04
Trace improvements

jodzga:
Tracing improvements:
  - Changed the way tasks fusion is reflected in trace so that it is  more intuitive and better reflect performance of individual transformations
  - Added class name of lambdas passed to functional API to default task descriptions

### DIFF
--- a/src/com/linkedin/parseq/FusionTask.java
+++ b/src/com/linkedin/parseq/FusionTask.java
@@ -22,13 +22,12 @@ import com.linkedin.parseq.trace.TraceBuilder;
 
 
 /**
+ * This is internal class and it should not be extended or used outside ParSeq.
+ * It may change at any time in a backwards incopmpatible way.
  *
- * @author jodzga
- *
- * @param <S>
- * @param <T>
+ * @author Jaroslaw Odzga (jodzga@linkedin.com)
  */
-public class FusionTask<S, T> extends BaseTask<T> {
+class FusionTask<S, T> extends BaseTask<T> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(FusionTask.class);
 
@@ -41,10 +40,9 @@ public class FusionTask<S, T> extends BaseTask<T> {
   private final ShallowTraceBuilder _predecessor;
 
   private FusionTask(final String desc, final Task<S> task, final Promise<S> source,
-      final Consumer3<FusionTraceContext, Promise<S>, Settable<T>> propagator, final boolean systemHidden,
+      final Consumer3<FusionTraceContext, Promise<S>, Settable<T>> propagator,
       final ShallowTraceBuilder predecessor) {
     super(desc);
-    _shallowTraceBuilder.setSystemHidden(systemHidden);
     _propagator = completing(propagator);
     _task = task;
     _source = source;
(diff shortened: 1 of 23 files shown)


src/com/linkedin/parseq/FusionTask.java
```diff
@@ -22,33 +22,61 @@
...
 
   private static final Continuations CONTINUATIONS = new Continuations();
 
+  /* Encapsulates transformation from source to target, includes all predecessors' transformations, can't be null */
   private final Consumer3<FusionTraceContext, Promise<S>, Settable<T>> _propagator;
-  private final Task<S> _task;
+
+  /* Asynchronous task that will complete source promise, can be null in which case propagator must not depend on source */
+  private final Task<S> _asyncTask;
+
```
zhenkai: It's a bit confusing why we need both a source task that will complete the source promise and the source promise itself. Can you add some text explaining why both of them are needed?

src/com/linkedin/parseq/FusionTask.java
```diff
@@ -22,33 +22,61 @@
...
-    _predecessor = predecessor;
+    _predecessor = null;
+  }
+
+  private <R> FusionTask(final String desc, final FusionTask<S, R> predecessor,
+      final PromisePropagator<R, T> propagator) {
+    super(desc);
+    _asyncTask = predecessor._asyncTask;
+    _source = predecessor._source;
+    _predecessor = predecessor.getShallowTraceBuilder();
```
zhenkai: maybe rename _predecessor to _predecessorShallowTraceBuilder... it's kind of long though

src/com/linkedin/parseq/FusionTask.java
```diff
@@ -57,56 +85,144 @@ private void trnasitionToDone(final FusionTraceContext traceContext) {
...
   private Consumer3<FusionTraceContext, Promise<S>, Settable<T>> completing(
       final Consumer3<FusionTraceContext, Promise<S>, Settable<T>> propagator) {
     return (traceContext, src, dest) -> {
       final SettablePromise<T> settable = FusionTask.this.getSettableDelegate();
 
-      if (traceContext.getTrigger() != null && traceContext.getTrigger().getId().equals(getId())) {
-        //simply propagate result because the BaseTask's code
-        //handles completing the task
+      if (isPropagationInitiator(traceContext)) {
+        /* BaseTask's code handles completing the parent task
```
zhenkai: Just to be sure, parent task means the source task & surrogate task means the fusion tasks that starts the propagation, right?

src/com/linkedin/parseq/FusionTraceContext.java
```diff
@@ -1,23 +1,55 @@
...
 
-  public ShallowTraceBuilder getTrigger() {
-    return _trigger;
+  public ShallowTraceBuilder getSurrogate() {
+    return _surrogate;
   }
 
-  public Context getContext() {
-    return _context;
+  public void createSurrogate() {
```
zhenkai: why do we need this method? there should always be a surrogate, no? I can imagine in some cases surrogate would be identical to propagationInitiator

zhenkai: why can't this be put into constructor? all the info needed is supplied at construction time

src/com/linkedin/parseq/FusionTask.java
```diff
@@ -248,27 +328,27 @@ protected void propagate(final FusionTraceContext traceContext, final SettablePr
...
     } else {
-      final Task<T> propagationTask = Task.async("fusion", ctx -> {
+      /* There is async predecessor, need to run it first */
+      _shallowTraceBuilder.setName("async fused");
+      _shallowTraceBuilder.setSystemHidden(true);
+      /* PropagationTask will actually run propagation */
+      final Task<T> propagationTask = Task.async(baseName, ctx -> {
         final SettablePromise<T> fusionResult = Promises.settable();
-        FusionTraceContext traceContext = new FusionTraceContext(ctx, FusionTask.this.getShallowTraceBuilder());
+        FusionTraceContext traceContext = new FusionTraceContext(ctx, FusionTask.this.getShallowTraceBuilder(), baseName);
```
zhenkai: This line & line 334 are basically the same, right? Or is the ctx different from context?

src/com/linkedin/parseq/FusionTask.java
jodzga: This was necessary at one of the previous versions but it is not anymore. I'll remove _source. Thanks for pointing this out.

src/com/linkedin/parseq/FusionTask.java
jodzga: It is mouthful but more descriptive, will change.

src/com/linkedin/parseq/FusionTraceContext.java
jodzga: Because surrogate is set only when there are at least 2 transformation. This complication is necessary to avoid wrapping single transformation with "fuse" parent.

src/com/linkedin/parseq/FusionTask.java
jodzga: They are different.

PR #66, opened 2015-11-05
Refactor Continuations implementation to simplify it a bit

zhenkai:
Use two stack to do post-order traversal of the execution tree.
ant test passed.

### DIFF
--- a/src/com/linkedin/parseq/internal/Continuations.java
+++ b/src/com/linkedin/parseq/internal/Continuations.java
@@ -1,7 +1,8 @@
 package com.linkedin.parseq.internal;
 
-import java.util.Stack;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 
 /**
  * This class allows running the following code structure:
@@ -43,16 +44,16 @@ public class Continuations {
   private static final class Continuation {
     // contains sibling actions in reverse order of submission
     // sibling actions are actions submitted by the same parent action
-    private Stack<Runnable> _siblingActions = new Stack<>();
+    private final Deque<Runnable> _siblingActions = new ArrayDeque<>();
 
     /** contains actions for execution in post-order
      *  actions with larger depth at the top;
      *  for sibling actions with the same depth, the first submitted is at the top
      */
-    private Stack<Runnable> _postOrderExecutionStack = new Stack<>();
+    private final Deque<Runnable> _postOrderExecutionStack = new ArrayDeque<>();
 
     private void submit(final Runnable action) {
-      if (_postOrderExecutionStack.empty()) {
+      if (_postOrderExecutionStack.size() == 0) {
         // we are at the root level of a call tree
         // this branch contains main loop responsible for
         // executing all actions
@@ -60,7 +61,7 @@ public class Continuations {
         loop();
       } else {
         // another child action added by the current action
-        _siblingActions.add(action);
+        _siblingActions.push(action);
       }
     }
 
@@ -77,10 +78,10 @@ public class Continuations {
           // currentAction could have submitted a few children actions, so we pop them out from
           // _siblingActions & push them into _postOrderExecutionStack, resulting in the desired
           // post-order execution order
-          while (!_siblingActions.empty()) {
+          while (_siblingActions.size() > 0) {
             _postOrderExecutionStack.push(_siblingActions.pop());
           }
-        } while (!_postOrderExecutionStack.empty());
+        } while (_postOrderExecutionStack.size() > 0);
       } finally {
         // maintain invariants
         _postOrderExecutionStack.clear();
--- a/src/com/linkedin/parseq/internal/Continuations.java
+++ b/src/com/linkedin/parseq/internal/Continuations.java
@@ -1,7 +1,6 @@
 package com.linkedin.parseq.internal;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
+import java.util.Stack;
 
 
 /**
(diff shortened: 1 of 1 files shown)


src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -1,7 +1,6 @@
 package com.linkedin.parseq.internal;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
+import java.util.Stack;
```
jodzga: Stack is not optimal data structure here because it is unnecessarily synchronized.

src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -32,10 +32,7 @@
   private final ThreadLocal<Continuation> CONTINUATION = new ThreadLocal<Continuation>() {
     @Override
     protected Continuation initialValue() {
-      Continuation cont = new Continuation();
```
jodzga: This is nice change :-)

jodzga:
Some refactoring is very nice e.g. moving initialization to constructor, additional documentation.
Change in the actual implementation is minor and to me is not much cleaner but I suspect this is very subjective.
My major objection is the use of java.util.Stack.

src/com/linkedin/parseq/internal/Continuations.java
zhenkai: you're right! I didn't check the javadoc :)

zhenkai:
The actual impl  is very similar. Just trying to make it a bit obvious :)

src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -44,51 +42,54 @@ public void submit(final Runnable action) {
...
-      //  - _scheduled has one action
+      //  - _siblingActions is empty
+      //  - _postOrderExecutionStack has one action (root)
       try {
         do {
-          final Runnable next = _scheduled.pollFirst();
-          next.run();
-          while (!_active.isEmpty()) {
-            _scheduled.addFirst(_active.pollLast());
+          Runnable currentAction = _postOrderExecutionStack.peek();
```
angxu: why not do #pop here?

zhenkai: If we do pop here, actions submitted by the parent action will not go into _siblingActions (see new line 56).

angxu: I see. Thanks!

src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -16,8 +16,9 @@
  * such that {@code actionX} can throw exception or recursively call {@code method}
  * multiple times, without worry about stack overflow.
  * <p>
- * The guarantee is that actions are called in the same order than recursive approch
- * would have called them.
+ * The guarantee is that actions are called in the same order than recursive approach
+ * would have called them. To put it in another way, this class guarantees to traversal the
+ * execution tree in post-order.
```
jodzga: The traversal is pre-order, isn't it?

src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -44,51 +42,54 @@ public void submit(final Runnable action) {
...
-    // variable to cache empty deque instance
-    private Deque<Runnable> _inactive;
-
-    // contains actions scheduled for execution
-    private Deque<Runnable> _scheduled;
+    /** contains actions for execution in post-order
+     *  actions with larger depth at the top;
+     *  for sibling actions with the same depth, the first submitted is at the top
+     */
+    private final Deque<Runnable> _postOrderExecutionStack = new ArrayDeque<>();
```
jodzga: Rename to _preOrderExecutionStack.

src/com/linkedin/parseq/internal/Continuations.java
```diff
@@ -44,51 +42,54 @@ public void submit(final Runnable action) {
...
 
     private void submit(final Runnable action) {
-      if (_active == null) {
+      if (_postOrderExecutionStack.size() == 0) {
```
jodzga: It would be even cleaner to have a nullable variable _activeAction that is set in submit() and nulled out in the finally block in the loop(). This way condition here can change to:
if (_activeAction == null)

and loop can further simplify to just do _postOrderExecutionStack.pop() instead of peek/pop combination.

zhenkai: _inLoop looks good :)

src/com/linkedin/parseq/internal/Continuations.java
zhenkai: Yeah, since root runnable is executed first, it should be pre-order. 
So there can be some difference between Continuations and the normal recursive call in terms of order. 

In recursive call this is post-order: doWork of root is invoked last.
```
void foo(Node node) {
  foo(node.left);
  foo(node.right);
  doWork(node);
}
```
Using continuations this is pre-order: doWork of root is invoked first.
```
void foo(Node node) {
  CONTINUATIONS.submit(() -> foo(node.left));
  CONTINUATIONS.submit(() -> foo(node.right));
  doWork(node);
}
```
There is also mentioning of "recursivePostOrder" in TestContinuations, which adds a bit more confusion....

jodzga: Yeah, this class does not claim to be general purpose replacement of recursion - see description in class Javadoc.
The recursivePostOrder in TestContinuations refers to the test case, not the implementation of Continuation class.
The Continuation class implementation is using pre-order traversal.

zhenkai: That's the confusing part... I thought it was testing post-order traversal implementation.

zhenkai:
changes already included in #67

PR #68, opened 2016-01-25
Initial commit for Parseq zookeeper client

angxu:
A thin wrapper around vanilla zookeeper client to facilitate the usage of Parseq. Also implemented a distributed lock based on http://zookeeper.apache.org/doc/trunk/recipes.html#Shared+Locks.


contrib/parseq-zk-client/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+            <groupId>com.linkedin.pegasus</groupId>
+            <artifactId>d2</artifactId>
+            <version>2.6.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.linkedin.pegasus</groupId>
+            <artifactId>r2-core</artifactId>
+            <version>2.6.0</version>
+        </dependency>
```
zhenkai: pegasus dependency seems unnecessary

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/PlanLocal.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+/**
+ * A {@link Map} that is shared across {@link Task}s belong to the same plan.
+ *
+ * @author Ang Xu
+ */
+public final class PlanLocal {
+
+  private static final ConcurrentMap<Long, Map<String, Object>> _planLocalMap = new ConcurrentHashMap<>();
```
zhenkai: why is the type <String, Object> for map?

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/Reaper.java
```diff
@@ -0,0 +1,104 @@
...
+  private final ExecutorService _executor;
+
+  private final ReaperRunnable _runnable = new ReaperRunnable();
+
+  public Reaper(Engine engine) {
+    this(engine, Executors.newSingleThreadExecutor(new NamedThreadFactory("ZKClient Reaper")));
+  }
+
+  public Reaper(Engine engine, ExecutorService executor) {
+    _queue = new ArrayBlockingQueue<>(10);
```
zhenkai: submit will throw if queue is full. is 10 large enough?
at least make it configurable.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/Reaper.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public void shutdown() {
+      _runnable = false;
+    }
+  }
+
+  @FunctionalInterface
+  public interface Zombie {
+    Task<Void> cleanUp();
```
zhenkai: javadoc to make clear the contract
looks like cleanUp need to return shareableTask or new task every time from above code (reuse zombie)

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/Reaper.java
```diff
@@ -0,0 +1,104 @@
...
+import com.linkedin.r2.util.NamedThreadFactory;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A reaper thread which is responsible for removing the zombie znode
```
zhenkai: why do we need a dedicated reaper thread? can we just use task & parseq engine to do the removal of node? i.e. when you want to remove a node, just submit a task
Of course, without the use of blocking queue

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
```diff
@@ -0,0 +1,434 @@
...
+      }
+    }
+    return promise;
+  }
+
+  /**
+   * Shuts down the zookeeper connection.
+   * @throws InterruptedException
+   */
+  public void shutdown() throws InterruptedException {
```
zhenkai: return Promise just to symmetric with start()...

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
```diff
@@ -0,0 +1,434 @@
...
+   * @param ops operations to execute.
+   * @param executor {@code Executor} that will be used to run the operations.
+   * @return task to execute multiple operations.
+   */
+  public Task<List<OpResult>> multi(Iterable<Op> ops, Executor executor) {
+    return Task.blocking(() -> _zkClient.multi(ops), executor);
+  }
+
+  /**
+   * Returns task that will wait for the given {@link KeeperState} to fulfil.
```
zhenkai: s/fulfil/fulfill

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
```diff
@@ -0,0 +1,434 @@
...
+
+  public Task<String> ensurePathExists(String path) {
+    return exists(path).flatMap(stat -> {
+      if (!stat.isPresent()) {
+        String parent = path.substring(0, path.lastIndexOf('/'));
+        if (parent.isEmpty()) { // parent is root
+          return create(path, null, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
+        } else {
+          return ensurePathExists(parent)
+              .flatMap(unused -> create(path, null, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
```
zhenkai: why use flatMap instead of andThen?

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKUtil.java
```diff
@@ -0,0 +1,107 @@
...
+
+
+/**
+ * @author Ang Xu
+ */
+public final class ZKUtil {
+
+  private ZKUtil() {}
+
+  public static String noramlizeZKPath(String basePath, String child) {
```
zhenkai: typo

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/PlanLocal.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+
+/**
+ * A {@link Map} that is shared across {@link Task}s belong to the same plan.
+ *
+ * @author Ang Xu
+ */
+public final class PlanLocal {
```
zhenkai: a more general question is why does the operations return task when the operations are not async in nature? It seems to be a wrong place to create task.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/recipes/ZKLock.java
```diff
@@ -0,0 +1,237 @@
...
+            // we already acquire the lock, add count only.
+            lockInternal._lockCount++;
+            return Task.value(lockInternal._lockNode);
+          } else {
+            // try acquire.
+            return acquire(deadline);
+          }
+        })
+        /* run the given task with toTry() */
+        .flatMap(unused -> task).toTry()
```
zhenkai: I already asked this question, but why not andThen?

contrib/parseq-zk-client/pom.xml
angxu: I'm using the NamedThreadFactory from r2...

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/recipes/ZKLock.java
```diff
@@ -0,0 +1,237 @@
...
+        /* Step 2: Call getChildren( ) on the lock node without setting the watch flag */
+        _zkClient.getChildren(_lockPath)
+            .flatMap(children -> {
+              children = children.stream().map(child -> ZKUtil.noramlizeZKPath(_lockPath, child))
+                  .collect(Collectors.toList());
+              String node = ZKUtil.findNodeWithNextLowestSN(children, lockNode);
+              if (node.equals(lockNode)) {
+                /* Step 3: If the pathname created in step 1 has the lowest sequence number suffix,
+                 * the client has the lock and the client exits the protocol. */
+                return PlanLocal.put(getPlanLocalKey(), new LockInternal(lockNode)).map(unused -> lockNode);
```
zhenkai: the value part is weird...
it seems we just want to put lockNode, but why coupling with instantiation.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/PlanLocal.java
angxu: The outer map stores the mapping between planId and its planLocals which is yet another map.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
angxu: But it's a blocking call anyway unless I submit the shutdown tasks to another executor, which seems overkilled.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
angxu: The javadoc of andThen(Task t) says that "Task passed in as a parameter will run even if the previous task completes fails", which is not what we want. Here we don't want to run the create task if ensurePathExists fails.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/recipes/ZKLock.java
angxu: ditto.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/PlanLocal.java
angxu: Because the operations need to be deferred until after the plan has been submitted and started. Doing it at task composition time defects the purpose of this class.

zhenkai: I don't quite understand. If you do the operation inside some task created elsewhere, the operation won't be executed util the plan got started.

contrib/parseq-zk-client/pom.xml
zhenkai: There is also d2.
Even for r2, I don't think it's worth pulling it in just for NamedThreadFactory, which is a trivial one.

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/ZKClient.java
zhenkai: That's a surprise to me! especially when there is another andThen which needs a consumer and the consumer apparently only would be called if the task succeeded.

using flatMap this way seems to suggest we need a new method...

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/PlanLocal.java
angxu: Yeah, you're right. But then I guess the API needs to take the Context as a parameter and you would have to convert every task which uses this API to async Task.

zhenkai: From the usages in this RB, it seems to me it's unnecessary to create tasks here. Some of the flatMap "unused" hacks are no longer needed if we don't create task here.

We can discuss offline

contrib/parseq-zk-client/pom.xml
angxu: Okay, I'll copy it over then. I'd keep d2 however because it's just a test dependency

contrib/parseq-zk-client/src/main/java/com/linkedin/parseq/zk/client/Reaper.java
angxu: Great point!

PR #69, opened 2016-02-10
Task based batching strategy

yahro:


contrib/parseq-batching/README.md
```diff
@@ -146,3 +146,23 @@ Finally we need to define main API for our ```ParSeqPersonClient```:
 ```batchable()``` method is declared by a ```BatchingStrategy``` and returns a task that cooperates with a batching strategy to performa a batchable operation.
 
 Source code for above example can be found [here](https://github.com/linkedin/parseq/blob/master/contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/domain/ParSeqPersonClient.java).
+
+Task-based BatchingStrategy
+===========================
+
+```BatchingStrategy``` API is intended to be used when integrating asynchronous API (e.g. based on callbacks or ```CompletableFuture``` ) with parseq. It is not convenient to use when we have an existing parseq API. In those cases we can use ```TaskBasedBatchingStrategy```.
+
+```TaskBasedBatchingStrategy``` class has 3 type parameters:
```
stevenheidel: This whole ~~Review Board~~ Pull Request looks awesome overall!

Honestly the only thing I don't like is the naming of the strategy. `TaskBasedBatchingStrategy` seems incredibly verbose. I would prefer just `TaskBatchingStrategy`. Just like in rest.li we have `AssociationResourceTaskTemplate` not `AssociationResourceTaskBasedTemplate`.

So I'm suggesting renaming to `TaskBatchingStrategy` and `SimpleTaskBatchingStrategy`.

yahro: Yeah, sounds much better.

stevenheidel:
Ship it! :+1:

PR #74, opened 2016-03-08
Cleaned up BaseEngineTest

jodzga:
Added trace logging to runAndWaitException.
Added variations of runAndWaitException.


angxu:
LGTM.

PR #75, opened 2016-03-14
Fix a bug in TestZKClient

angxu:
Fix a bug in TestZKClient where we didn't wait for the zookeeper connection to be established completely.


jodzga:
Can you also add a log4j configuration file to the src/test/resources e.g:
https://github.com/linkedin/parseq/blob/master/test/log4j.properties

angxu:
Oops..

PR #83, opened 2016-06-27
Adding PlanCompletionListener

angxu:
Adding PlanCompletionListener which gets invoked when all the tasks(including sideEffect task) start by this plan have been finished.

### DIFF
--- a/src/com/linkedin/parseq/internal/PlanContext.java
+++ b/src/com/linkedin/parseq/internal/PlanContext.java
@@ -3,6 +3,7 @@ package com.linkedin.parseq.internal;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 
+import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.ILoggerFactory;
 import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


src/com/linkedin/parseq/internal/PlanContext.java
```diff
@@ -91,6 +126,43 @@ public String getPlanClass() {
...
+    }
+    return null;
+  }
+
+  /**
+   * Decrements the pending count by 1. Invokes {@link #_planCompletionListener}
+   * if there is no more pending (sub-)plans.
+   */
+  private void done() {
+    int pending;
```
jodzga: replace with decrementAndGet()

jodzga:
Could you add unit tests?
Does this approach work when there are tasks scheduled for execution using Context.createTimer()?

angxu:
Sure. 
I feel there is no need to handle timer task explicitly. If a timer task was fired and caused the plan to complete, it'll be captured just like the normal task. Otherwise, if the plan is completed before the timer task fires, then it is either already cancelled or is about to be cancelled, which doesn't make much difference from a tracing standpoint.

PR #87, opened 2016-08-05
Configurable parseq restli client with batching

jodzga:


angxu:
LGTM other than the typo of "keysSize".

PR #88, opened 2016-08-10
Make EngineBuilder#setPlanDeactivationListener backward compatible with v2.3.*

angxu:
Make EngineBuilder#setPlanDeactivationListener backward compatible with v2.3.*


FranklinYinanDing:
Looks like it should return EngineBuilder(but can't because of backward incompatible). Can you please add a TODO to remind for next major version? Thanks! Ship it!

PR #90, opened 2016-09-06
Added limit on max concurrent plans to be executed by the Engine

jodzga:
Engine.run(...) methods are blocking now when the max concurrent plans limit is reached. For backwards compatibility default value for this limit is Integer.MAX_VALUE, what essentially means "unbounded capacity".
Added non blocking and time-bounded blocking tryRun(...) methods.

### DIFF
--- a/src/com/linkedin/parseq/Engine.java
+++ b/src/com/linkedin/parseq/Engine.java
@@ -19,6 +19,7 @@ package com.linkedin.parseq;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -46,6 +47,9 @@ public class Engine {
   public static final String MAX_RELATIONSHIPS_PER_TRACE = "_MaxRelationshipsPerTrace_";
   private static final int DEFUALT_MAX_RELATIONSHIPS_PER_TRACE = 4096;
 
+  public static final String MAX_CONCURRENT_PLANS = "_MaxConcurrentPlans_";
+  private static final int DEFUALT_MAX_CONCURRENT_PLANS = Integer.MAX_VALUE;
+
   private static final State INIT = new State(StateName.RUN, 0);
   private static final State TERMINATED = new State(StateName.TERMINATED, 0);
   private static final Logger LOG = LoggerFactory.getLogger(LOGGER_BASE);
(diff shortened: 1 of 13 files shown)


src/com/linkedin/parseq/Engine.java
```diff
@@ -103,6 +96,37 @@
...
+    } else {
+      _maxRelationshipsPerTrace = DEFUALT_MAX_RELATIONSHIPS_PER_TRACE;
+    }
+
+    if (_properties.containsKey(MAX_CONCURRENT_PLANS)) {
+      _maxConcurrentPlans = (Integer) getProperty(MAX_CONCURRENT_PLANS);
+    } else {
+      _maxConcurrentPlans = DEFUALT_MAX_CONCURRENT_PLANS;
+    }
+    _concurrentPlans = new Semaphore(_maxConcurrentPlans);
```
angxu: Do we want FIFO guarantee on the semaphore? It seems counter intuitive to me that the plan submitted first may fail with a timeout but the plan submitted later can go through.

src/com/linkedin/parseq/Engine.java
```diff
@@ -113,25 +137,127 @@
...
+   * For the sake of backwards compatibility default value for a {@value #MAX_CONCURRENT_PLANS} is
+   * {@value #DEFUALT_MAX_CONCURRENT_PLANS} which essentially means "unbounded capacity".
+   *
+   * @param task the task to run
+   * @param planClass string that identifies a "class" of the Plan. Plan class ends up in a ParSeq
+   * Trace and can be used to group traces into "classes" when traces are statistically analyzed.
+   */
+  public void run(final Task<?> task, final String planClass) {
+    try {
+      _concurrentPlans.acquire();
```
angxu: It feels like this can easily cause lock contention problem. Can we move this feature into a custom PlanCompletionListener?

src/com/linkedin/parseq/Engine.java
jodzga: I would leave it as an implementation detail. I see two scenarios:
1) In setting like kafka stream consumer the submitting thread is expecting to block and in this case there will be no timeout
2) In async settings, if Engine is above it's capacity some task submissions will potentially time out but it is not clear to me why one would expect FIFO semantics, e.g. is FIFO better than other strategy, if so, why?

src/com/linkedin/parseq/Engine.java
jodzga: Can you expand on where the lock contention could occur? Do you mean scenario when Engine does not reach it's capacity limit?

src/com/linkedin/parseq/Engine.java
angxu: I didn't think of any real life scenario but was just out of curiosity. After a second thought, I think you're right that people expecting FIFO behavior can and probably should submit their tasks one after another using blocking API, so it doesn't matter how semaphore behaves internally.

src/com/linkedin/parseq/Engine.java
jodzga: Non-fair Semaphore.aquire() is implemented as an int CAS and falls back into locking only when number of permits is exhausted, so I don't think there is lock contention risk as long as Engine does not reach it's capacity limit. Once Engine reaches it's capacity limit we do expect locking and blocking.

angxu: got it!

PR #91, opened 2016-09-07
Changed behavior of Engine.run() to not block but throw an IllegalStateException

jodzga:
Changed behavior of Engine.run() to not block but throw an
IllegalStateException once Engine's capacity has been exceeded.

Some ParSeq users follow an anti-pattern of not building plans by
composition but instead are executing new plans using Engine.run() even
from within another task. Changing semantics to blocking would cause
unrecoverable deadlock. It is better to throw an exception in this case
because it can/should be handled by code that executes Engine.run().


angxu:
Ship it!

PR #92, opened 2016-09-14
Parseq retry policy implementation.

extesy:
Support for flexible retry policies, based on http://zman.io/atmos/ concepts.

---
This change is [Reviewable](https://reviewable.io/reviews/linkedin/parseq/92)


extesy:
@jodzga Looks like an unrelated transient failure in travis, not related to this PR.

extesy:
Pinging @angxu for code review.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  protected Promise<? extends T> run(Context context) {
+    _startedAt = System.currentTimeMillis();
+    Task<T> task = wrap(0);
```
angxu: So the attempt is zero-indexed? I think we need some explicit documentation about this.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryFailureException.java
```diff
@@ -0,0 +1,17 @@
...
+public class RetryFailureException extends Exception {
+  private static final long serialVersionUID = 0L;
+
+  public RetryFailureException() {
+    super();
+  }
+
+  public RetryFailureException(final String message) {
+    super(message);
+  }
```
angxu: Could you add a new constructor which takes a Throwable and calls super(Throwable t) so that we don't lose the stack trace when there is a fatal error?

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/backoff/RandomizedBackoff.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * A policy that randomizes the result of another policy by adding a random duration in the specified range.
+ *
+ * @param <T> Type of a task result, same as in a base policy.
+ *
+ * @author Oleg Anashkin (oleg.anashkin@gmail.com)
+ */
+public class RandomizedBackoff<T> implements BackoffPolicy<T> {
+  protected final Random _random = new Random();
```
angxu: ThreadLocalRandom is better:)

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/LogEventsWithJava.java
```diff
@@ -0,0 +1,84 @@
...
+
+
+/**
+ * An event monitor that formats and logs events using the `java.util.logging` framework.
+ *
+ * @param <T> Type of a task result, used for strongly typed processing of outcomes.
+ *
+ * @author Oleg Anashkin (oleg.anashkin@gmail.com)
+ */
+public class LogEventsWithJava<T> extends LogEvents<T, Level> {
```
angxu: This class doesn't seem necessary since we already have implementation against SLF4J API.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/NeverTerminate.java
```diff
@@ -0,0 +1,23 @@
+package com.linkedin.parseq.retry.termination;
+
+
+/**
+ * A termination policy that never signals for termination.
+ *
+ * @author Oleg Anashkin (oleg.anashkin@gmail.com)
+ */
+public class NeverTerminate implements TerminationPolicy {
```
angxu: What's the use case of this policy? It scares me by its name...

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/AbstractRetryPolicy.java
```diff
@@ -0,0 +1,146 @@
...
+
+
+/**
+ * A base class for policies that enable customizable retries for arbitrary parseq tasks.
+ *
+ * @param <T> Type of a task result, used for strongly typed processing of outcomes.
+ *
+ * @author Oleg Anashkin (oleg.anashkin@gmail.com)
+ */
+public abstract class AbstractRetryPolicy<T> {
```
extesy: Sure, will do.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * @return True if this classification will not record information about a failure.
+   */
+  public boolean isSilent() {
+    return _isSilent;
+  }
+
+  protected static boolean nonFatal(Throwable throwable) {
+    return throwable instanceof StackOverflowError || !(throwable instanceof VirtualMachineError || throwable instanceof ThreadDeath
```
extesy: I copied this implementation from Scala's standard NonFatal function. I don't know details, but I guess they have reasons for classifying exceptions like this.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java
```diff
@@ -0,0 +1,35 @@
...
+  public ErrorClassification getStatus() {
+    return _status;
+  }
+
+  /**
+   * Associate the error classification with this result. It controls the behavior of retry operation.
+   *
+   * @param status Error classification.
+   */
+  public ResultClassification setStatus(ErrorClassification status) {
```
extesy: Because that would need three unacceptable values: UNACCEPTABLE_FATAL, UNACCEPTABLE_RECOVERABLE and UNACCEPTABLE_SILENTLY_RECOVERABLE, essentially tightly coupling it with ErrorClassification enums. I didn't want to do that. I'll see what I can do about avoiding global effect.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * A parseq task wrapper that supports arbitrary retry policies.
+   *
+   * @param taskFunction A task generator function. It will receive a zero-based attempt number as a parameter.
+   * @param policy Retry policy that will control this task's behavior.
+   */
+  public RetriableTask(Function<Integer, Task<T>> taskFunction, AbstractRetryPolicy<T> policy)
+  {
+    this("operation", taskFunction, policy);
```
extesy: This is the name of operation, not the task. It's used in logging to construct messages like "attempt 2 of <operationName> failed because of <reason>", so default "operation" name reads much better in this context than "withRetry".

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java
extesy: I do have explicit documentation about this, please check javadocs on the publicly available functions. Did I miss it somewhere?

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryFailureException.java
extesy: Currently it's only invoked in the context when no parent Throwable is available, so this new constructor would be useless. I could add it, but why?

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/backoff/RandomizedBackoff.java
extesy: Sure.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/LogEventsWithJava.java
extesy: This is for completeness sake - not every developer in the world uses slf4j, and also for example purposes.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/NeverTerminate.java
extesy: Someone might needed it for simple side projects. I'll add a warning to javadoc discouraging usage of this policy.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java
extesy: If we treat all errors as fatal by default, there won't be any retries by default either, which is not very useful. Here fatal errors are more like system-fatal.

angxu: Sorry, but Errors I mean: https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html, but all the exceptions.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryFailureException.java
angxu: Yeah, I should also say that you can use the new constructor at line 160 in RetriableTask with the fatal error you get. That way when we look at the RetryFailureException from a failed parseq plan, we get see not only the error message but also the actual stack trace.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java
angxu: Okay. _status should be volatile btw.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java
angxu: Ah, I misunderstood it.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java
angxu: Nope. I missed that.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryFailureException.java
extesy: But outcome.getError() is of `Throwable` type so it will already include stack trace of the actual task error. It's a bad idea to wrap it in extra `RetryFailureException` because that would essentially hide common task failures like `RestliException` and would always require unpacking `RetryFailureException` by callers who wish to use retry wrapper.

Existing `new RetryFailureException` on line 160 is on the right side of ternary operator, so it means there is no `Throwable` available from the task and there's nothing to pass to that new constructor.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/AbstractRetryPolicy.java
extesy: @angxu What do you think about merging `RetryPolicy` and `AbstractRetryPolicy`? I don't foresee any use cases for extending either of them, so having only one class `RetryPolicy` as both builder and container simplifies things.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryFailureException.java
angxu: You are right that there is no stack trace to pass in. For some reason I missed the ternary operator. You can ignore the comment.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java
extesy: I replaced enum with a class but made sure that its usage syntax is exactly the same.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/TerminationPolicy.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  static TerminationPolicy requireEither(TerminationPolicy first, TerminationPolicy second) {
+    return new RequireEither(first, second);
+  }
+
+  /**
+   * A termination policy that always signals for termination.
+   */
+  static TerminationPolicy alwaysTerminate() {
+    return new AlwaysTerminate();
```
angxu: This could be a singleton

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/TerminationPolicy.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * A termination policy that never signals for termination.
+   * WARNING: Please think twice before using this policy, it could cause a deadlock in your application.
+   */
+  static TerminationPolicy neverTerminate() {
+    return new NeverTerminate();
```
angxu: ditto.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/AbstractRetryPolicy.java
angxu: The best practice we've been using so far is to define both an interface and a builder class - the interface will have all the getters and builder will have all the setters. You can either have a separate impl class or combine it with your builder, but make sure you don't leak the constructor. That way you are free to make incremental changes to your interface in the future without breaking backward compatibility.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/TerminationPolicy.java
extesy: Yes, it could be, but the implementation wouldn't look nice. Since `TerminationPolicy` is an interface, it can't have private fields. So either I make ALWAYS_TERMINATE static field (inconsistent naming), or I implement `AlwaysTerminate.instance()` function. Does it really matter, though? These two policies are going to be used very rarely, I doubt they would ever be instantiated more than once in any application.

angxu: I haven't tried it out myself, but can you do `return (attempts, nextAttemptAt) -> true;`? 
In anyways, it's not a big issue and I'm fine if you leave it as it is.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/AbstractRetryPolicy.java
extesy: Ok, done.

contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/termination/TerminationPolicy.java
extesy: Yeah, it works, I like it. Thanks for suggestion.

extesy:
Review status: 0 of 35 files reviewed at latest revision, 11 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 57 at r3](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRdeaF1X40o9It5Xe41-r3-57:-KRkLOw9xLsHCNazwPNf:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/9fed94b1cff67fe338d6131ed5e4d74b662bf062/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L57)):*
<details><summary><i>Previously, angxu wrote…</i></summary>
> Sorry, but Errors I mean: https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html, but all the exceptions.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Reviewed 6 of 35 files at r1, 1 of 5 files at r2, 5 of 11 files at r4, 1 of 2 files at r5, 3 of 4 files at r6.
Review status: 16 of 35 files reviewed at latest revision, 19 unresolved discussions.

---

*[contrib/parseq-batching/pom.xml, line 112 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlF9Il1vDtIM2nCGiQ:-KRlF9Im8sGClfFp6_Qq:b-ucfknh) ([raw file](https://github.com/linkedin/parseq/blob/1608fb0ab70b8d68aa061bd9ac71defce027dc0f/contrib/parseq-batching/pom.xml#L112)):*
> ```XML
>     </scm>
>     <description>
>       Provides convenient API for creatinig automatically batched tasks
> ```

Thanks!

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 57 at r3](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRdeaF1X40o9It5Xe41-r3-57:-KRgjdE2cwx4d7cMF8UJ:b-3f4os3) ([raw file](https://github.com/linkedin/parseq/blob/9fed94b1cff67fe338d6131ed5e4d74b662bf062/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L57)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> Done.
</details>
I think making all Errors 'FATAL' is too strict, e.g. IOError doesn't seem like it should be 'FATAL'. I would just pick explicit short list of errors enumerated in Scala's NonFatal.
Also, CancellationException should not be "RETRYABLE".

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 11 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlG3896CxFggsNxymS:-KRlG3896CxFggsNxymT:bwrcdup) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L11)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public enum ErrorClassification {
> ```

This is a bit pedantic but perhaps name "RETRYABLE" instead "RECOVERABLE" fits better here?
I would simplify this to have no variables and just two instances: RETRYABLE and UNRETRYABLE.  I would avoid using name 'FATAL' because it is already used in Scala and might be confusing. 
I would also drop SILENTLY_RECOVERABLE. Do you have a real-life use case when you would use it? Why one would choose it?

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java, line 9 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlHmfyjva9GMm9qkk6:-KRlHmfyjva9GMm9qkk7:b-dd60b) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java#L9)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public final class ResultClassification {
> ```

I wonder if ResultClassification is even needed. Wouldn't it be simpler and more "regular" to make tasks with "UNACCEPTABLE" results fail and then use ErrorClassification on it? That would simplify the design IMO.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 25 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlItyV7YX0JJD-DXcK:-KRlItyV7YX0JJD-DXcL:b-dzbph6) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L25)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public class RetriableTask<T> extends BaseTask<T> {
> ```

This is minor but it would be nice to not extend BaseTask and instead use Task.async(). The reason is that extending BaseTask makes it a bit harder to refactor it in the future.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 55 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlJnkUBXPwlv4u_E5j:-KRlJnkUBXPwlv4u_E5k:b-30ywls) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L55)):*
> ```Java
>    * @param <U> Type of a task result, used for strongly typed processing of outcomes.
>    */
>   static <U> RetryPolicyBuilder<U> attempts(int attempts) {
> ```

Perhaps it would be a good idea to add "backoff" parameter to those helper functions and provide one more backoff policy "none" or something similar that would mean "no backoff" to make it possible to implement retries without backoff. The reason is that it might not be super obvious and missed that retries without any backoff may be dangerous (e.g. can overload downstream service when retrying RPC).

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLREwfO5hpZ2pLJbF:-KRlLREwfO5hpZ2pLJbG:bhxsvf6) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L17)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public final class RetryPolicyBuilder<T> implements RetryPolicy<T> {
> ```

build() method is missing in this builder?

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 57 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLVmnq5FlDzIMDIeV:-KRlLVmnq5FlDzIMDIeW:b-3f9d86) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L57)):*
> ```Java
>    * @param errorClassifier The classifier for errors raised during retry operations.
>    */
>   static public <U> RetryPolicyBuilder<U> create(String name, TerminationPolicy terminationPolicy, BackoffPolicy<U> backoffPolicy, EventMonitor<U> eventMonitor,
> ```

Remove this method to make future evolution of this builder easier (e.g. adding new field).

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlDnyQRtpw2Tq6hmqc:-KRlDnyQRtpw2Tq6hmqd:b-hfqual) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L17)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public interface EventMonitor<T> {
> ```

Considering that parseq trace will contain all information about the retries why is event monitoring necessary? If it is just about logging, wouldn't it be simpler to add debug level logging using slf4j just lie we do in other parts of parseq?
Do you have a specific use case in mind for event monitoring? Or perhaps explain why tracing is not adequate mechanism?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Review status: 16 of 35 files reviewed at latest revision, 20 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 18 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlQwXRXluF0-u0Zuns:-KRlQwXSEYRMtP8zLtyD:byzaz05) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L18)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public interface RetryPolicy<T> {
> ```

Perhaps remove "Policy" suffix in names e.g. RetryPolicy -> Retry, BackoffPolicy -> Backoff, withRetryPolicy() -> withRetry(). This would make names a bit shorter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 16 of 35 files reviewed at latest revision, 20 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 57 at r3](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRdeaF1X40o9It5Xe41-r3-57:-KRnwR6pWNGBE6y4YAGo:bibt6iw) ([raw file](https://github.com/linkedin/parseq/blob/9fed94b1cff67fe338d6131ed5e4d74b662bf062/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L57)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I think making all Errors 'FATAL' is too strict, e.g. IOError doesn't seem like it should be 'FATAL'. I would just pick explicit short list of errors enumerated in Scala's NonFatal.
> Also, CancellationException should not be "RETRYABLE".
</details>
Reverted back to Scala list with CancellationException. 

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 11 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlG3896CxFggsNxymS:-KRnvYrYMxSEm-kBn4_N:bo1ztbx) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L11)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> This is a bit pedantic but perhaps name "RETRYABLE" instead "RECOVERABLE" fits better here?
> I would simplify this to have no variables and just two instances: RETRYABLE and UNRETRYABLE.  I would avoid using name 'FATAL' because it is already used in Scala and might be confusing. 
> I would also drop SILENTLY_RECOVERABLE. Do you have a real-life use case when you would use it? Why one would choose it?
</details>
The use case for SILENTLY_RECOVERABLE is the usage of LimitDuration policy which can potentially generate a lot of retries. In combination with logging event monitor that could generate a lot of logging noise, while the only actionable log would be a single "aborted" error or nothing if the operation eventually succeeded. So I'd prefer to keep it.

As for RETRYABLE/UNRETRYABLE - maybe it's just me, but "recoverable" sounds better than "retriable/retryable", and it's also used much more often in industry: https://www.google.com/trends/explore?q=retriable,retryable,recoverable

I can rename it if you have a strong opinion about this. Please let me know.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java, line 9 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlHmfyjva9GMm9qkk6:-KRnwhVFQAhAwGSh_Qh1:brc4rr9) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java#L9)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I wonder if ResultClassification is even needed. Wouldn't it be simpler and more "regular" to make tasks with "UNACCEPTABLE" results fail and then use ErrorClassification on it? That would simplify the design IMO.
</details>
It might make sense for LinkedIn + RestLi, but ParSeq is an open source project which is not tightly coupled with restli and could potentially be used in other contexts. Imagine a task that uses HttpClient from http://hc.apache.org/httpclient-3.x/tutorial.html - it won't throw any exception. The only way to mark a request as failed would be to examine its result (status code).

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 25 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlItyV7YX0JJD-DXcK:-KRo2CLhvc3H17sBWeZV:b3io1z9) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L25)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> This is minor but it would be nice to not extend BaseTask and instead use Task.async(). The reason is that extending BaseTask makes it a bit harder to refactor it in the future.
</details>
BaseTask's description says "An abstract base class that can be used to build implementations of Task" and since RetriableTask is actually a task, that's why it inherits from BaseTask.  If I inherit RetriableTask from Task instead, then I'd have to reimplement parts of BaseTask anyways and copy/pasting is bad.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 18 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlQwXRXluF0-u0Zuns:-KRnyfQckoZKlVtqBGpU:bbz8093) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L18)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Perhaps remove "Policy" suffix in names e.g. RetryPolicy -> Retry, BackoffPolicy -> Backoff, withRetryPolicy() -> withRetry(). This would make names a bit shorter.
</details>
I would disagree with this. Policy in this context is a synonym for strategy. This interface doesn't really do any retry by itself (it's done by RetriableTask), it only defines the behavior and controls the flow inside RetriableTask.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 55 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlJnkUBXPwlv4u_E5j:-KRo00NYH6QneSAmDjWW:b-4gkv6o) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L55)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Perhaps it would be a good idea to add "backoff" parameter to those helper functions and provide one more backoff policy "none" or something similar that would mean "no backoff" to make it possible to implement retries without backoff. The reason is that it might not be super obvious and missed that retries without any backoff may be dangerous (e.g. can overload downstream service when retrying RPC).
</details>
Done. I left functions without backoff for use cases where having backoff doesn't make any sense. Not any task involves network call. It might be as simple as showing a UI dialog, so why force backoff in such case?

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLREwfO5hpZ2pLJbF:-KRo0JccOF2JLCMdZjZt:b-i88ras) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L17)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> build() method is missing in this builder?
</details>
It's not needed here because RetryPolicyBuilder implements RetryPolicy. I used this simplified builder pattern so that it's possible to easily modify policies created with this helper using "RetryPolicy.<Void>attempts(3).setErrorClassifier(errorClassifier)" pattern. If I used a real decouple builder pattern then do adjust precreated policy I'd either have to not use a helper and construct full policy myself, or modify helpers to return a builder instead of policy so that all callers will have to include .build() call. Both options look bad to me, so I went with the pattern that simplifies the usage syntax. Thoughts?

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 57 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLVmnq5FlDzIMDIeV:-KRo7gTc247yIufoFGI6:b-q64pkt) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L57)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Remove this method to make future evolution of this builder easier (e.g. adding new field).
</details>
See above.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlDnyQRtpw2Tq6hmqc:-KRo17PfhqZw8CbYRHhf:b-4cc9hv) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L17)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Considering that parseq trace will contain all information about the retries why is event monitoring necessary? If it is just about logging, wouldn't it be simpler to add debug level logging using slf4j just lie we do in other parts of parseq?
> Do you have a specific use case in mind for event monitoring? Or perhaps explain why tracing is not adequate mechanism?
</details>
Few reasons: 1) I didn't want to force slf4j to the users, 2) it's not necessarily about logging, there can be other behavior implemented, for example notifying user that the task is taking longer than usual so please wait.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Reviewed 1 of 2 files at r8.
Review status: 15 of 35 files reviewed at latest revision, 20 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java, line 11 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlG3896CxFggsNxymS:-KRodxw9Lt1RarBthQGT:b-5zp4d1) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ErrorClassification.java#L11)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> The use case for SILENTLY_RECOVERABLE is the usage of LimitDuration policy which can potentially generate a lot of retries. In combination with logging event monitor that could generate a lot of logging noise, while the only actionable log would be a single "aborted" error or nothing if the operation eventually succeeded. So I'd prefer to keep it.
> 
> As for RETRYABLE/UNRETRYABLE - maybe it's just me, but "recoverable" sounds better than "retriable/retryable", and it's also used much more often in industry: https://www.google.com/trends/explore?q=retriable,retryable,recoverable
> 
> I can rename it if you have a strong opinion about this. Please let me know.
</details>
I don't have a strong opinion about naming but let's come back to the SILENTLY_RECOVERABLE once we agree on what we do with events monitoring.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java, line 9 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlHmfyjva9GMm9qkk6:-KRoSYY8eeqw-S_KUUWA:bsbpb0s) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java#L9)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> It might make sense for LinkedIn + RestLi, but ParSeq is an open source project which is not tightly coupled with restli and could potentially be used in other contexts. Imagine a task that uses HttpClient from http://hc.apache.org/httpclient-3.x/tutorial.html - it won't throw any exception. The only way to mark a request as failed would be to examine its result (status code).
</details>
I realize my comment might not be clear enough. Let me clarify.

Let's say that we have httpTask that will always complete successfully but for some results we would like to issue a retry.

With current proposal one wouldcreate a RetryPolicy with custom ResultClassifier e.g. (pseudo-code):

retryPolicy = RetryPolicyBuilder
                            .setBackoffPolicy(...)
                            .setResultsPolicy(<<retry-if-http-code != 200>>)
                            .build()
mytask = httpTask.withRetryPolicy(retryPolicy)

What I am saying is that an equivalent solution would be:

retryPolicy = RetryPolicyBuilder
                            .setBackoffPolicy(...)
                            .build()

mytask  = httpTask.map(<<fail-if-http-code != 200>>).withRetryPolicy(retryPolicy)
                   
This way RetryPolicy only retries when there was a failure and if result is 'not acceptable' then one might consider that task has failed.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 25 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlItyV7YX0JJD-DXcK:-KRoaubrqjMQq_5UAgh8:b-r00u92) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L25)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> BaseTask's description says "An abstract base class that can be used to build implementations of Task" and since RetriableTask is actually a task, that's why it inherits from BaseTask.  If I inherit RetriableTask from Task instead, then I'd have to reimplement parts of BaseTask anyways and copy/pasting is bad.
</details>
Aaah, I need to fix the javadoc for a BaseTask.
If you think about it what BaseTask gives you is a way to get a Task implementation if you provide function: context -> Promise (by implementing an abstract run(Context):Promise method.
You get exactly the same functionality by calling Task.async(Function<Context, Promise<T>>).

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 18 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlQwXRXluF0-u0Zuns:-KRoboQwbTwQtPwNVqTC:b32w0x) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L18)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> I would disagree with this. Policy in this context is a synonym for strategy. This interface doesn't really do any retry by itself (it's done by RetriableTask), it only defines the behavior and controls the flow inside RetriableTask.
</details>
OK

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 55 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlJnkUBXPwlv4u_E5j:-KRofsFMkisrFJfY08e0:b-za5aag) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L55)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> Done. I left functions without backoff for use cases where having backoff doesn't make any sense. Not any task involves network call. It might be as simple as showing a UI dialog, so why force backoff in such case?
</details>
I understand that in some cases one can retry immediately but the problem is that in some cases this is a terrible idea and it might be easily misused. We should try to 'hint' users that they should consider backoff policy. To simplify usage  for cases where no backoff is needed we provide noBackoff.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLREwfO5hpZ2pLJbF:-KRocSrB0_eemrw36kNW:b-rrz2if) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L17)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> It's not needed here because RetryPolicyBuilder implements RetryPolicy. I used this simplified builder pattern so that it's possible to easily modify policies created with this helper using "RetryPolicy.<Void>attempts(3).setErrorClassifier(errorClassifier)" pattern. If I used a real decouple builder pattern then do adjust precreated policy I'd either have to not use a helper and construct full policy myself, or modify helpers to return a builder instead of policy so that all callers will have to include .build() call. Both options look bad to me, so I went with the pattern that simplifies the usage syntax. Thoughts?
</details>
Oh, I missed that RetryPolicyBuilder implements RetryPolicy. This is bad idea IMO because it makes RetryPolicy mutable. I'd rather see a common builder pattern with mutable builder and immutable returned object.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlDnyQRtpw2Tq6hmqc:-KRodLbCFkcUENp_iK_1:b-ygpbda) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L17)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> Few reasons: 1) I didn't want to force slf4j to the users, 2) it's not necessarily about logging, there can be other behavior implemented, for example notifying user that the task is taking longer than usual so please wait.
</details>
1) Parseq is already depending on slf4j so this ship has sailed ;-) 
2) One can do the same by either adding a PromiseListener or notify a user in a retry task. It is a quite large bit of code and I would like to better understand why are we adding it, what are alternatives etc. Once we ship it we'll have to maintain it forever.
Are you planning to use it in your project? If so, how?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 12 of 36 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java, line 9 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlHmfyjva9GMm9qkk6:-KRos_a4_sNZcg_u4Ou9:bbu4lrb) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java#L9)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I realize my comment might not be clear enough. Let me clarify.
> 
> Let's say that we have httpTask that will always complete successfully but for some results we would like to issue a retry.
> 
> With current proposal one wouldcreate a RetryPolicy with custom ResultClassifier e.g. (pseudo-code):
> 
> retryPolicy = RetryPolicyBuilder
>                             .setBackoffPolicy(...)
>                             .setResultsPolicy(<<retry-if-http-code != 200>>)
>                             .build()
> mytask = httpTask.withRetryPolicy(retryPolicy)
> 
> What I am saying is that an equivalent solution would be:
> 
> retryPolicy = RetryPolicyBuilder
>                             .setBackoffPolicy(...)
>                             .build()
> 
> mytask  = httpTask.map(<<fail-if-http-code != 200>>).withRetryPolicy(retryPolicy)
>                    
> This way RetryPolicy only retries when there was a failure and if result is 'not acceptable' then one might consider that task has failed.
> 
</details>
Ok, I see now. Yes, this should be possible to do.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 25 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlItyV7YX0JJD-DXcK:-KRoqaDQoKtJjfTZdvNr:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L25)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Aaah, I need to fix the javadoc for a BaseTask.
> If you think about it what BaseTask gives you is a way to get a Task implementation if you provide function: context -> Promise (by implementing an abstract run(Context):Promise method.
> You get exactly the same functionality by calling Task.async(Function<Context, Promise<T>>).
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 55 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlJnkUBXPwlv4u_E5j:-KRorjxTT_HesV4Him26:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L55)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I understand that in some cases one can retry immediately but the problem is that in some cases this is a terrible idea and it might be easily misused. We should try to 'hint' users that they should consider backoff policy. To simplify usage  for cases where no backoff is needed we provide noBackoff.
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlLREwfO5hpZ2pLJbF:-KRorlj-yhMczwQ2hEDJ:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L17)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Oh, I missed that RetryPolicyBuilder implements RetryPolicy. This is bad idea IMO because it makes RetryPolicy mutable. I'd rather see a common builder pattern with mutable builder and immutable returned object.
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 17 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlDnyQRtpw2Tq6hmqc:-KRorwR2nfwKDjejCEN2:b-u98j18) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L17)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> 1) Parseq is already depending on slf4j so this ship has sailed ;-) 
> 2) One can do the same by either adding a PromiseListener or notify a user in a retry task. It is a quite large bit of code and I would like to better understand why are we adding it, what are alternatives etc. Once we ship it we'll have to maintain it forever.
> Are you planning to use it in your project? If so, how?
</details>
I used the Scala version of it to do logging in the context of the caller function and not ParSeq. When you pass the same logger that is used in the function itself, it's much easier to filter relevant errors by logger name, rather than digging through everything that parseq logs for the entire application.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 3 of 34 files reviewed at latest revision, 19 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java, line 9 at r7](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRlHmfyjva9GMm9qkk6:-KRp1LvPbVzsDdfPZvRs:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/0c3499d894f66c05043352f344fb6ce60ddc4f1b/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/ResultClassification.java#L9)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> Ok, I see now. Yes, this should be possible to do.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Review status: 3 of 34 files reviewed at latest revision, 17 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 96 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq1iJL6c1T9qrXeGpd:-KRq1iJL6c1T9qrXeGpe:b-f95nu) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L96)):*
> ```Java
>   public static <U> Task<U> withRetryPolicy(String name, RetryPolicy policy, Function<Integer, Task<U>> taskFunction) {
>     RetriableTask<U> retriableTask = new RetriableTask<>(name, policy, taskFunction);
>     return Task.async(retriableTask::run);
> ```

Add task name to show up in a trace e.g.
return Task.async("retriableTask", retriableTask::run);

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 103 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq488X2lOub6Nf8xz1:-KRq488X2lOub6Nf8xz2:bnxycup) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L103)):*
> ```Java
>     Task<T> task = _taskFunction.apply(attempt);
> 
>     return Task.async(_policy.getName(), context -> {
> ```

For better debugability add attempt number to the task's name e.g.:

return Task.async(_policy.getName() + ", attempt: " + attempt, context -> {

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 48 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRpNwWItLM97F56sWVy:-KRpNwWItLM97F56sWVz:baanavd) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L48)):*
> ```Java
>    * @param backoff The constant delay between retry attempts.
>    */
>   static RetryPolicy attempts(int attempts, long backoff) {
> ```

I was thinking more that BackoffPolicy would be a parameter's type.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 33 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq5FphsNVMdoArEHj_:-KRq5FphsNVMdoArEHja:bb6eqs6) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L33)):*
> ```Java
>     if (_retryPolicy.getName() == null) {
>       StringBuilder policyName = new StringBuilder("RetryPolicy");
>       String terminationPolicyName = _retryPolicy.getTerminationPolicy().getClass().getSimpleName();
> ```

It would be great to have even nicer names because it will help a lot in readability of traces. One solution might be:
- add getName(): String to TerminationPolicy and BackoffPolicy
- implementation could provide accurate name e.g. ConstantBackoff could be "constant backoff 100ms"
I think it would help a lot if backoff name had an actual time parameter in it

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyImpl.java, line 15 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq-n4OiZu8sN9D5vAS:-KRq-n4PELZtPkgNB4Oj:b2k4yij) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyImpl.java#L15)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public final class RetryPolicyImpl implements RetryPolicy {
> ```

Make this class package private so it can be refactored in the future without affecting users.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 13 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRqABpLOsDq5ntSkQWW:-KRqABpMxuFKzdiHxfch:b-hzgxgf) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L13)):*
> ```Java
>  * @author Oleg Anashkin (oleg.anashkin@gmail.com)
>  */
> public interface EventMonitor {
> ```

Sorry, I think I accidentally closed a discussion about EventMonitoring and don't know how to get back to it (I'm new to reviewable.io).
My understanding is that exactly the same information in already in parseq trace. Can you comment on why it can't be used for retries or what information is missing?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 3 of 34 files reviewed at latest revision, 17 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 13 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRqABpLOsDq5ntSkQWW:-KRqUsV5GSbRqhdwjo5P:b-kl7z1l) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L13)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Sorry, I think I accidentally closed a discussion about EventMonitoring and don't know how to get back to it (I'm new to reviewable.io).
> My understanding is that exactly the same information in already in parseq trace. Can you comment on why it can't be used for retries or what information is missing?
</details>
I used the Scala version of it to do logging in the context of the caller function and not ParSeq. When you pass the same logger that is used in the function itself, it's much easier to filter relevant errors by logger name, rather than digging through everything that parseq logs for the entire application.

Some information might be already there in parseq trace, but I needed log messages with logger name matching the caller function, not internal parseq class.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Review status: 3 of 34 files reviewed at latest revision, 17 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 13 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRqABpLOsDq5ntSkQWW:-KRtAHNDciJUn_vLklOs:b-qzmhmp) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L13)):*
<details><summary><i>Previously, extesy (Oleg Anashkin) wrote…</i></summary>
> I used the Scala version of it to do logging in the context of the caller function and not ParSeq. When you pass the same logger that is used in the function itself, it's much easier to filter relevant errors by logger name, rather than digging through everything that parseq logs for the entire application.
> 
> Some information might be already there in parseq trace, but I needed log messages with logger name matching the caller function, not internal parseq class.
</details>
I see. Wouldn't you achieve similar debugability level if you added slf4j logging statements in relevant retry-related classes? When debugging you could then enable debug logging for com.linkedin.parseq.retry level and you would see only retry-related statements.
The only gap I see is that you would not be able to parameterize it with custom logger name but I would trade off this inconvenience over added complexity of com.linkedin.parseq.retry.* classes - you basically introduced a little monitoring subsystem ;-). Using slf4j would also be more familiar to all users in general. It would also cut down the codebase.
One more disadvantage of current implementation is that monitoring concern leaked out to the retry logic:  ErrorClassification enum has SILENTLY_RECOVERABLE and _isSilent variable.
Could we drop entire com.linkedin.parseq.retry.monitor package and instead add bunch of slf4j debug log statements? Or would it create a gap in functionality in your opinion?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 3 of 34 files reviewed at latest revision, 17 unresolved discussions.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 96 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq1iJL6c1T9qrXeGpd:-KS8E82nFTMvjApam_qS:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L96)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Add task name to show up in a trace e.g.
> return Task.async("retriableTask", retriableTask::run);
> 
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java, line 103 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq488X2lOub6Nf8xz1:-KS8EIzx1pp4Ce7L3Hco:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetriableTask.java#L103)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> For better debugability add attempt number to the task's name e.g.:
> 
> return Task.async(_policy.getName() + ", attempt: " + attempt, context -> {
> 
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java, line 48 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRpNwWItLM97F56sWVy:-KS8EMifVBYM8pO_fIQX:ba0hhbl) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java#L48)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I was thinking more that BackoffPolicy would be a parameter's type.
</details>
I think that's too heavy for a supposedly simple helper function. When forcing caller to provide so much details, the benefit of using a helper over a builder is minimal.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java, line 33 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq5FphsNVMdoArEHj_:-KS8EdNQAnRmDq_ivKW7:b-ym5dky) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyBuilder.java#L33)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> It would be great to have even nicer names because it will help a lot in readability of traces. One solution might be:
> - add getName(): String to TerminationPolicy and BackoffPolicy
> - implementation could provide accurate name e.g. ConstantBackoff could be "constant backoff 100ms"
> I think it would help a lot if backoff name had an actual time parameter in it
</details>
I think that current approach is "good enough". Having each policy construct its name in a generic format overcomplicates design (see your own event monitor comments).

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyImpl.java, line 15 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRq-n4OiZu8sN9D5vAS:-KS8EvusJYr4DqkcJm_S:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/RetryPolicyImpl.java#L15)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> Make this class package private so it can be refactored in the future without affecting users.
</details>
Done.

---

*[contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java, line 13 at r9](https://reviewable.io:443/reviews/linkedin/parseq/92#-KRqABpLOsDq5ntSkQWW:-KS8GZf5nzHjsPybC_lw:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/9bfa00d7dbb4d55da4dad01041919ad3f5ac9a7c/contrib/parseq-retry/src/main/java/com/linkedin/parseq/retry/monitor/EventMonitor.java#L13)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> I see. Wouldn't you achieve similar debugability level if you added slf4j logging statements in relevant retry-related classes? When debugging you could then enable debug logging for com.linkedin.parseq.retry level and you would see only retry-related statements.
> The only gap I see is that you would not be able to parameterize it with custom logger name but I would trade off this inconvenience over added complexity of com.linkedin.parseq.retry.* classes - you basically introduced a little monitoring subsystem ;-). Using slf4j would also be more familiar to all users in general. It would also cut down the codebase.
> One more disadvantage of current implementation is that monitoring concern leaked out to the retry logic:  ErrorClassification enum has SILENTLY_RECOVERABLE and _isSilent variable.
> Could we drop entire com.linkedin.parseq.retry.monitor package and instead add bunch of slf4j debug log statements? Or would it create a gap in functionality in your opinion?
> 
> 
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Review status: 2 of 23 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[contrib/parseq-retry/README.md, line 75 at r10](https://reviewable.io:443/reviews/linkedin/parseq/92#-KS8hNN-X5C-oa_on7wU:-KS8hNN0ACjDfaHtgGIx:b-iiwyb8) ([raw file](https://github.com/linkedin/parseq/blob/ba9a1a6045f513a064a7a1a9614a773b81819923/contrib/parseq-retry/README.md#L75)):*
> ```Markdown
> ================
> 
> Simple retry policy from the examples above would retry failed tasks immediately. To implement some delay between retries, the number of milliseconds should be passed to the policy:
> ```

One thing worth clarifying and not immediately obvious is that backoff time is counted from the completion of the task e.g. if I make HTTP requests with constant backoff of 1s it doesn't mean requests will be made every 1s - if it takes 1s to get HTTP response then requests will be made every 2s.

---

*[contrib/parseq-retry/src/test/java/com/linkedin/parseq/retry/TestBackoffPolicy.java, line 4 at r10](https://reviewable.io:443/reviews/linkedin/parseq/92#-KS8gkPLLoZ9iqGkTvYg:-KS8gkPLLoZ9iqGkTvYh:b88u9my) ([raw file](https://github.com/linkedin/parseq/blob/ba9a1a6045f513a064a7a1a9614a773b81819923/contrib/parseq-retry/src/test/java/com/linkedin/parseq/retry/TestBackoffPolicy.java#L4)):*
> ```Java
> 
> import com.linkedin.parseq.BaseEngineTest;
> import com.linkedin.parseq.function.Failure;
> ```

unused import

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

extesy:
Review status: 2 of 23 files reviewed at latest revision, 13 unresolved discussions.

---

*[contrib/parseq-retry/README.md, line 75 at r10](https://reviewable.io:443/reviews/linkedin/parseq/92#-KS8hNN-X5C-oa_on7wU:-KS8mCyIDsj6w6jQQ1jI:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/ba9a1a6045f513a064a7a1a9614a773b81819923/contrib/parseq-retry/README.md#L75)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> One thing worth clarifying and not immediately obvious is that backoff time is counted from the completion of the task e.g. if I make HTTP requests with constant backoff of 1s it doesn't mean requests will be made every 1s - if it takes 1s to get HTTP response then requests will be made every 2s.
</details>
Done.

---

*[contrib/parseq-retry/src/test/java/com/linkedin/parseq/retry/TestBackoffPolicy.java, line 4 at r10](https://reviewable.io:443/reviews/linkedin/parseq/92#-KS8gkPLLoZ9iqGkTvYg:-KS8mEZT7Vej3wG7sz8Z:b-896fix) ([raw file](https://github.com/linkedin/parseq/blob/ba9a1a6045f513a064a7a1a9614a773b81819923/contrib/parseq-retry/src/test/java/com/linkedin/parseq/retry/TestBackoffPolicy.java#L4)):*
<details><summary><i>Previously, jodzga wrote…</i></summary>
> unused import
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/92)*

jodzga:
Ship it!

PR #95, opened 2016-10-04
Allow user to wire custom TaskQueue implementation into SerialExecutor

angxu:
This patch should give us the ability to override the default FIFOPriorityQueue implementation in SerialExecutor.

---

This change is [Reviewable](https://reviewable.io/reviews/linkedin/parseq/95)

### DIFF
--- a/src/com/linkedin/parseq/internal/SerialExecutor.java
+++ b/src/com/linkedin/parseq/internal/SerialExecutor.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.parseq.internal;
 
-import com.linkedin.parseq.Priority;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -40,7 +39,7 @@ import java.util.concurrent.atomic.AtomicInteger;
  * @author Chris Pettitt (cpettitt@linkedin.com)
  * @author Jaroslaw Odzga (jodzga@linkedin.com)
  */
-public class SerialExecutor implements Executor {
+public class SerialExecutor {
 
   /*
    * Below is a proof and description of a mechanism that ensures that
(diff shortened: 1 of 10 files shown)


angxu:
Fixed the javadoc 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/linkedin/parseq/95#-:-KVbCTZav5dYD11FyKLj:b613ko0)*

src/com/linkedin/parseq/internal/SerialExecutor.java
```diff
@@ -140,12 +157,22 @@ public void run() {
...
+    void add(T value);
+    T poll();
+  }
+
   /*
    * Deactivation listener is notified when this executor finished executing a Runnable
    * and there are no other Runnables waiting in queue.
    * It is executed sequentially with respect to other Runnables executed by this Executor.
    */
-  static interface DeactivationListener {
```
jodzga: Why this is not static anymore?

angxu: Because nested interface is static by default and my IDE suggested me to remove the static keyword.

src/com/linkedin/parseq/internal/SerialExecutor.java
```diff
@@ -69,24 +70,40 @@
...
+    _queue = taskQueue;
     _deactivationListener = deactivationListener;
   }
 
   public void execute(final Runnable runnable) {
-    _queue.add(runnable);
+    if (runnable instanceof Prioritizable) {
+      _queue.add((PrioritizableRunnable)runnable);
+    } else {
+      _queue.add(new PrioritizableRunnable() {
```
jodzga: When does this wrapping happen?

src/com/linkedin/parseq/internal/SerialExecutor.java
jodzga: Interesting - this means that default methods can only access static methods of outer class.

src/com/linkedin/parseq/internal/SerialExecutor.java
jodzga: Can we drop this check then?

angxu: We could. But it means we'll have to do an unchecked type casting which can fail hard. Alternatively, we can drop the "implements Executor" declaration on SerialExecutor and enforce PrioritizableRunnable at method level. Note that SerialExecutor is already pretty tightly coupled with engine because we let user wire in PlanDeactivationListener and TaskQueue, so I don't see much value to keep the Executor interface.

jodzga: Good point. Can you drop "implements Executor"? I think this will clean up the implementation a bit.

PR #98, opened 2016-11-03
Unit test showing bug in batching shareable tasks.

jodzga:
Unit test showing bug caused by interaction of forking plans for side effects and batching.

The problem is that batching uses plan id as a key for aggregation but side effects fork plans what creates new plan ids and as a consequence batchable tasks never get executed.
This bug affects side effects in general and shareable() method because it uses runSideEffect()

---
This change is [Reviewable](https://reviewable.io/reviews/linkedin/parseq/98)


angxu:
fixed in 3001cc33bfe2a0d9841b16b568193dbca1025499

PR #107, opened 2017-02-10
fixing test in TestRequestContextProvider. 

ssodhanilinkedin:
The problem was some tests which were borrowed from TestParseqRestClient were not clearing contexts. Looks like these tests are already part of TestParSeqRestClient so removing them.


jodzga:
Ship it!

PR #112, opened 2017-02-23
Adding parseq-lambda-names contrib project which figures out better t…

ssodhanilinkedin:
…ask descriptions using ASM. In this commit, I just want to figure out source code location and integrate parseq and this contrib project using java SPI


contrib/parseq-lambda-names/README.md
```diff
@@ -0,0 +1,5 @@
+Parseq Lambda Names
+==========================
+
+This project aims to provide more meaningful descriptions for Parseq tasks. Using ASM, it tries to idenfity where lambda expression is defined in source code and also infer some details about its execution like function call within lambda expression with number of arguments.
```
jodzga: meaningful descriptions -> meaningful default descriptions

contrib/parseq-lambda-names/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+            <version>6.9.9</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
```
jodzga: This dependency should be removed or moved to "test" scope

contrib/parseq-lambda-names/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+    <version>2.6.8</version>
+
+    <dependencies>
+        <dependency>
+        <groupId>com.linkedin.parseq</groupId>
+            <artifactId>parseq</artifactId>
+            <version>2.6.8</version>
+            <!-- TODO: this needs to be improved -->
+            <scope>system</scope>
+            <systemPath>${basedir}/../../dist/parseq-${version}.jar</systemPath>
```
jodzga: Why is this needed?

src/com/linkedin/parseq/TaskDescriptorFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static ServiceLoader<TaskDescriptor> _loader;
+  private static TaskDescriptor _identifier;
+
+  static {
+    try {
+      _loader = ServiceLoader.load(TaskDescriptor.class);
+    } catch (SecurityException e) {
+      LOGGER.error("Unable to get ContextClassLoader for loading providers of TaskDescriptor", e.getMessage());
+    } catch (RuntimeException e) {
```
jodzga: Catch Throwable instead to log Errors as well.

src/com/linkedin/parseq/TaskDescriptor.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * An API to provide description for task.
+ *
+ * parseq-lambda-names provide an implementation for this interface using Java SPI. Any changes made to this interface
+ * might require a change to implementation in parseq-lambda-names.
+ *
+ * @author Siddharth Sodhani (ssodhani@linkedin.com)
+ */
+public interface TaskDescriptor {
```
jodzga: Do we need a new Interface here? Isn't it just Function<Class<?>, String>?

src/com/linkedin/parseq/TaskDescriptorFactory.java
```diff
@@ -0,0 +1,70 @@
...
+  static final Logger LOGGER = LoggerFactory.getLogger(TaskDescriptorFactory.class);
+
+  private static ServiceLoader<TaskDescriptor> _loader;
+  private static TaskDescriptor _identifier;
+
+  static {
+    try {
+      _loader = ServiceLoader.load(TaskDescriptor.class);
+    } catch (SecurityException e) {
+      LOGGER.error("Unable to get ContextClassLoader for loading providers of TaskDescriptor", e.getMessage());
```
jodzga: Are we sure that SecurityException  means problem with getting ContextClassLoader?

src/com/linkedin/parseq/TaskDescriptorFactory.java
```diff
@@ -0,0 +1,70 @@
...
+      return _identifier;
+    }
+
+    try {
+      Iterator<TaskDescriptor> identifiers = _loader.iterator();
+      if (identifiers.hasNext()) {
+        _identifier = identifiers.next();
+
+        if (identifiers.hasNext()) {
+          LOGGER.info("Found more than one provider for TaskDescriptor, picked first one");
```
jodzga: Log which one is used e.g. it's class name and the full list e.g. all class names to help debugging issues.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaClassLocator.java
```diff
@@ -0,0 +1,37 @@
+package com.linkedin.parseq.lambda;
+
+import org.objectweb.asm.ClassVisitor;
+
+
+/**
+ * This class visits all the classes and locates classes generated for Lambda expressions.
```
jodzga: This comment is not very clear to me e.g. what does it mean that it visits all classes.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
```diff
@@ -0,0 +1,82 @@
...
+import java.security.ProtectionDomain;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.Opcodes;
+
+
+public class ASMBasedTaskDescriptor implements TaskDescriptor {
```
jodzga: Add javadoc.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
```diff
@@ -0,0 +1,82 @@
...
+
+  static {
+    AgentLoader.loadAgentClass(Agent.class.getName(), null);
+  }
+
+  private static ConcurrentMap<String, String> _names = new ConcurrentHashMap<>();
```
jodzga: private static -> private static final

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public ASMBasedTaskDescriptor() {
```
jodzga: not needed?

contrib/parseq-lambda-names/README.md
ssodhanilinkedin: Done.

contrib/parseq-lambda-names/pom.xml
ssodhanilinkedin: Removed.

contrib/parseq-lambda-names/pom.xml
ssodhanilinkedin: Thats because TaskDescriptor is defined in parseq main project. ASMBasedTaskDescriptor implements that.

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: Not exactly but when there is security problems with getting ClassLoader.

https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html#getSystemClassLoader()

removing this catch clause as now catching throwable to cover all scenarios.

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: done.

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: I am not sure how would it work in JAVA SPI where in parseq-lambda-names has to provide implementation for this interface. Moreover I feel keeping as interface it looks more cleaner.

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaClassLocator.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
ssodhanilinkedin: removed

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/FindMethodCallAnalyzer.java
```diff
@@ -0,0 +1,179 @@
...
+                  field = methodInstr.name + Util.getArgumentsInformation(methodInstr.desc);
+                }
+              }
+            }
+          }
+
+          _inferredOperation = getInferredOperation(localVariables, field);
+        }
+      } catch (AnalyzerException e) {
+        //TODO: throw an exception that couldn't infer operation
```
jodzga: TODO

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/FindMethodCallAnalyzer.java
```diff
@@ -0,0 +1,179 @@
...
+        LOGGER.error("Unable to analyze class, could not infer operation", e.getMessage());
```
jodzga: We probably should log on DEBUG level because better lambda names are an optional improvement and we don't want to fill log with this kind of error messages if e.g. someone uses unsupported JVM.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaClassDescription.java
```diff
@@ -0,0 +1,39 @@
...
+    _sourcePointer = sourcePointer;
+    _inferredOperationOptional = Optional.of(inferredOperation);
+  }
+
+  String getDescription() {
+    StringBuilder builder = new StringBuilder();
+    if (_inferredOperationOptional.isPresent()) {
+      builder.append(_inferredOperationOptional.get()).append(" ");
+    } else {
+      //TODO: the best way to log this so that these could be looked into, may be log source pointer atleast to debug
```
jodzga: TODO

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaMethodVisitor.java
```diff
@@ -0,0 +1,103 @@
...
+  public void visitEnd() {
+    if (_containsSyntheticLambda) {
+      String classToVisit = _methodInsnOwner.replace('/', '.');
+      try {
+        SyntheticLambdaAnalyzer syntheticLambdaAnalyzer = new SyntheticLambdaAnalyzer(api, classToVisit, _methodInsnName);
+        ClassReader cr = new ClassReader(classToVisit);
+        cr.accept(syntheticLambdaAnalyzer, 0);
+        _inferredOperationConsumer.accept(new InferredOperation(_methodInsnOpcode, syntheticLambdaAnalyzer.getInferredOperation()));
+      } catch (IOException e) {
+        LOGGER.error("Unable to read class: " + classToVisit, e.getMessage());
```
jodzga: Change to DEBUG level

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaMethodVisitor.java
```diff
@@ -0,0 +1,103 @@
...
+        _inferredOperationConsumer.accept(new InferredOperation(Opcodes.INVOKESTATIC, functionName));
+      } else {
+        String classToVisit = _lambdaSourcePointer._className.replace('/', '.');
+        try {
+          FindMethodCallAnalyzer methodCallAnalyzer = new FindMethodCallAnalyzer(api, classToVisit, _lambdaSourcePointer, _methodInsnName);
+          ClassReader cr = new ClassReader(classToVisit);
+          cr.accept(methodCallAnalyzer, 0);
+          _inferredOperationConsumer.accept(new InferredOperation(_methodInsnOpcode, methodCallAnalyzer.getInferredOperation()));
+        } catch (IOException e) {
+          LOGGER.error("Unable to read class: " + classToVisit, e.getMessage());
```
jodzga: change to DEBUG level

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
```diff
@@ -0,0 +1,190 @@
...
+        case Opcodes.INVOKESPECIAL:
+        case Opcodes.INVOKESTATIC:
+          _methodInsnName = name;
+          _methodInsnOwner = owner;
+          _methodInsnDesc = desc;
+          _methodInsnOpcode = opcode;
+          break;
+        default:
+          //TODO: throw an exception that couldn't infer operation
+          LOGGER.debug("Unexpected opcode, falling back");
```
jodzga: TODO + logger level to DEBUG

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Override
+    public void visitEnd() {
+      try {
+        Analyzer analyzer = new Analyzer(new SourceInterpreter());
+        Frame[] frames = analyzer.analyze(_classToAnalyze, this);
+
+        int index = findMethodCall(this.instructions);
+        if (index == -1) {
+          LOGGER.error("Unable to find method call in instruction list, debug as this case is not expected");
```
jodzga: DEBUG

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
```diff
@@ -0,0 +1,190 @@
...
+                fieldDesc = methodInstr.name + Util.getArgumentsInformation(methodInstr.desc);
+              }
+            }
+          }
+        }
+
+        _inferredOperation = getInferredOperation(localVariables, fieldDesc);
+      } catch (AnalyzerException e) {
+        //TODO: throw an exception that couldn't infer operation
+        LOGGER.error("Unable to analyze class, could not infer operation", e.getMessage());
```
jodzga: TODO + DEBUG

src/com/linkedin/parseq/TaskDescriptorFactory.java
```diff
@@ -0,0 +1,69 @@
...
+  static final Logger LOGGER = LoggerFactory.getLogger(TaskDescriptorFactory.class);
+
+  private static ServiceLoader<TaskDescriptor> _loader;
+  private static TaskDescriptor _identifier;
+
+  static {
+    try {
+      _loader = ServiceLoader.load(TaskDescriptor.class);
+    } catch (Throwable e) {
+      LOGGER.error("Unable to load providers of TaskDescriptor", e.getMessage());
```
jodzga: Will there be error logged if there is no TaskDescriptor provided i.e. the current case?

contrib/parseq-lambda-names/README.md
```diff
@@ -0,0 +1,5 @@
+Parseq Lambda Names
+==========================
+
+This project aims to provide more meaningful default descriptions for Parseq tasks. Using ASM, it tries to locate where lambda expression is defined in source code and also infer some details about its execution like function call within lambda expression with number of arguments.
+
```
jodzga: Explain how this project is supposed to be used (include on a classpath) and that it is fragile and can break (generate different output) potentially between minor JVM versions.
Also, include info with what JVM versions this is working correctly.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/FindMethodCallAnalyzer.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/FindMethodCallAnalyzer.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaClassDescription.java
ssodhanilinkedin: removed

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaMethodVisitor.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaMethodVisitor.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
ssodhanilinkedin: done

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/SyntheticLambdaAnalyzer.java
ssodhanilinkedin: done

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: not at this point. it does lazy loading so we would get exception at line 44.

contrib/parseq-lambda-names/pom.xml
```diff
@@ -0,0 +1,158 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.linkedin.parseq</groupId>
+    <artifactId>parseq-lambda-names</artifactId>
+    <version>2.6.8</version>
+
```
jodzga: I think build should produce fat jar with all dependencies shadowed, but perhaps you'd rather do it in gradle instead?

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaClassLocator.java
```diff
@@ -0,0 +1,62 @@
...
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+
+
+/**
+ * An implementation of ASM ClassVisitor which analyses classes as they are loaded and identifies classes
+ * generated for Lamda expressions. In addition, it infers details such as source code location, function call within
+ * Lambda expression..
+ */
+public class LambdaClassLocator extends ClassVisitor {
```
jodzga: Change to package private, we don't want anyone to depend on this class.

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/LambdaMethodVisitor.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.IOException;
+import java.util.function.Consumer;
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class LambdaMethodVisitor extends MethodVisitor {
```
jodzga: Change to package private, we don't want anyone to depend on this class.

src/com/linkedin/parseq/TaskDescriptor.java
jodzga: I agree that keeping explicit interface seems cleaner.

src/com/linkedin/parseq/TaskDescriptorFactory.java
jodzga: In other words, if someone just bumps up ParSeq version and not include lambda-names artifact in a classpath then this error will not be logged?

src/com/linkedin/parseq/TaskDescriptorFactory.java
```diff
@@ -0,0 +1,69 @@
...
+      if (identifiers.hasNext()) {
+        _identifier = identifiers.next();
+        LOGGER.info("Using TaskDescriptor: " + _identifier.getClass());
+
+        while (identifiers.hasNext()) {
+          TaskDescriptor descriptor = identifiers.next();
+          LOGGER.debug("Discarding TaskDescriptor: {} as its not first one to be loaded: ", descriptor.getClass());
+        }
+      } else {
+        LOGGER.error("No provider found for TaskDescriptor, falling back to DefaultTaskDescriptor");
```
jodzga: This should be INFO level.

src/com/linkedin/parseq/TaskDescriptor.java
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Given class which could correspond to generated lambda expressions etc infer appropriate description for it
+   *
+   * @param clazz
+   * @return description for task
+   */
+  String getDescription(Class<?> clazz);
```
jodzga: One thing I am wondering if we could change this interface to:
```
String getDescription(String clazz);
```
The reason is that we might want to use TaskDescriptor in the future for purposes other than this. For example, it would be cool use it to "fix up" stack traces of exceptions in ParSeq. For this purpose it would be more convenient to use just String to avoid loading Class by name.

jodzga:
It would be good to add a TestUnrecognizedLambda with few cases where lambda name was not recognized and how it would look like e.g. it should still have a class name and line number.

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: This would not be logged. but "No provider found for TaskDescriptor, falling back to DefaultTaskDescriptor: this would be logged.

contrib/parseq-lambda-names/pom.xml
ssodhanilinkedin: done.

contrib/parseq-lambda-names/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+                                    <pattern>org.objectweb</pattern>
+                                    <shadedPattern>parseq.org.objectweb</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.ea</pattern>
+                                    <shadedPattern>parseq.com.ea</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>org.slf4j</pattern>
+                                    <shadedPattern>parseq.org.slf4j</shadedPattern>
```
ssodhanilinkedin: done.

src/com/linkedin/parseq/Task.java
```diff
@@ -192,7 +192,7 @@
    * @see #map(String, Function1)
    */
   default <R> Task<R> map(final Function1<? super T, ? extends R> func) {
-    return map("map: " + func.getClass().getName(), func);
+    return map("map: " + func.getClass().getName(), func);
```
ssodhanilinkedin: fixed

ssodhanilinkedin:
added TestUnrecognizedLambda with 2 tests for now. Need to add more.

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: I am not sure because in that case conversion from Class to String would have to be done by user (parseq main in this case) and String is too generic.

src/com/linkedin/parseq/TaskDescriptorFactory.java
ssodhanilinkedin: done.

contrib/parseq-lambda-names/README.md
ssodhanilinkedin: added.

src/com/linkedin/parseq/TaskDescriptor.java
jodzga: It is true that the input to this method eventually will always be a Class and it is true for Task interface where we want to generate better task description.
What I am thinking is that this could be useful in potentially different scenarios. One example is a stack trace. Currently stack trace contains lambda class name. It might be possible to improve it by substituting it with description provided by this class. In this scenario we would have only a class name as a String and converting it to Class for the purpose of this interface could be time consuming.
Do we use the Class object passed to this method in any way except for calling toString() on it?

contrib/parseq-batching/pom.xml
```diff
@@ -4,13 +4,13 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.linkedin.parseq</groupId>
     <artifactId>parseq-batching</artifactId>
-    <version>2.6.8</version>
+    <version>2.6.9-SNAPSHOT</version>
```
jodzga: Version number change should be removed before merging with master.

contrib/parseq-lambda-names/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,185 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jodzga: What is this file for?

contrib/parseq-lambda-names/pom.xml
```diff
@@ -0,0 +1,192 @@
...
+            <groupId>com.ea.agentloader</groupId>
+            <artifactId>ea-agent-loader</artifactId>
+            <version>1.0.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm-all</artifactId>
+            <version>5.1</version>
+        </dependency>
+        <dependency>
```
jodzga: I am wondering if we should remove dependency on slf4j? The problem is that some slf4j binding is still required and it needs to be available on System Class Loader what might not always be the case. I am wondering if should replace all LOG.xxx with System.out?

contrib/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
```diff
@@ -0,0 +1,124 @@
...
+ * within lambda.
+ */
+public class ASMBasedTaskDescriptor implements TaskDescriptor, Serializable {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ASMBasedTaskDescriptor.class);
+
+  private static ConcurrentMap<String, String> _names = new ConcurrentHashMap<>();
+
+  static {
+    if (ASMBasedTaskDescriptor.Agent.class.getClassLoader() != ClassLoader.getSystemClassLoader()) {
```
jodzga: Please document this non-trivial code

contrib/parseq-lambda-names/src/test/java/com/linkedin/parseq/lambda/TestInterface.java
```diff
@@ -0,0 +1,100 @@
...
+    assertNameMatch("impl::abstractFunction", "testAbstractFunctionReferenceOnInterface", CLASSNAME, description.get().toString());
+  }
+
+  @Test
+  public void testFunctionInvocationOnInterface() {
+    SampleImplementation impl = new SampleImplementation();
+    Optional<String> description = getDescriptionForFunction(s -> impl.interfaceFunction(s));
+    assertTrue(description.isPresent());
+    if (JDK_VERSION.equals(JDK_1_8_0_72)) {
+      assertNameMatch("(impl,s) -> interfaceFunction(_)", "testFunctionInvocationOnInterface", CLASSNAME, description.get().toString());
```
jodzga: This doesn't seam right. "s -> interfaceFunction(_)" looks acceptable to me but "(impl,s) -> interfaceFunction(_)" suggests that it is a lambda shape that takes two arguments while it takes only one argument. This might be confusing if one would like to locate it in a source code.

If it is not possible to distinguish between the two then perhaps it would be better to completely ditch left hand side for now. We can always try to improve it in the future versions.
If there is a significant number of cases for which left hand side is reliable then we could replace left hand side where we are unsure with "(?) ->". If it is difficult to even detect where we are unsure then we could remove left hand side completely for now and revisit it in the future.

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: Done.

contrib/parseq-lambda-names/dependency-reduced-pom.xml
ssodhanilinkedin: deleted

PR #118, opened 2017-03-09
Adding parseq-lambda-names contrib project which figures out better t…

ssodhanilinkedin:
…ask descriptions using ASM. In this commit, I just want to figure out source code location and integrate parseq and this contrib project using java SPI


src/com/linkedin/parseq/Task.java
```diff
@@ -64,6 +64,8 @@
 public interface Task<T> extends Promise<T>, Cancellable {
   static final Logger LOGGER = LoggerFactory.getLogger(Task.class);
 
+  static final TaskDescriptor _taskDescriptor = TaskDescriptorFactory.getTaskDescriptor();
```
jodzga: Some usages were nor replaced e.g. line 399 (before changes):
```
  /**
   * Equivalent to {@code andThen("andThen", consumer)}.
   * @see #andThen(String, Consumer1)
   */
  default Task<T> andThen(final Consumer1<? super T> consumer) {
    return andThen("andThen: " + consumer.getClass().getName(), consumer);
  }
```

Can you verify other usages?

src/com/linkedin/parseq/TaskDescriptor.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * parseq-lambda-names provide an implementation for this interface using Java SPI. Any changes made to this interface
+ * might require a change to implementation in parseq-lambda-names.
+ *
+ * @author Siddharth Sodhani (ssodhani@linkedin.com)
+ */
+public interface TaskDescriptor {
+
+  /**
+   * Give class name which could correspond to generated lambda expressions etc infer appropriate description for it
```
jodzga: What if lambda class was not found e.g. if I pass a gibberish to this function? What is it supposed to return?
Perhaps it would be better to change signature to `Optional<String> getDescription(String className);`?

src/com/linkedin/parseq/TaskDescriptor.java
```diff
@@ -0,0 +1,20 @@
+package com.linkedin.parseq;
+
+/**
+ * An API to provide description for task.
+ *
+ * parseq-lambda-names provide an implementation for this interface using Java SPI. Any changes made to this interface
+ * might require a change to implementation in parseq-lambda-names.
+ *
```
jodzga: Does implementation need to be thread safe? Can you add this info to the javadoc?

src/com/linkedin/parseq/Task.java
ssodhanilinkedin: done.

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: added.

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: actually we just return className as default, I updated javadoc to reflect that. Thoughts?

jodzga:
In the initial prototype there was a  mechanism to detect unknown bytecode patterns and return a fallback result in that case. For example, the following test:
```
  @Test
  public void testBlock() {
    Optional<String> lambdaName = getLambdaName(str -> {
      if (str.length() > 0) {
        return str.trim();
      } else {
        return str;
      }
    });
    assertTrue(lambdaName.isPresent());
    System.out.println(lambdaName);
  }
```
would print `Optional[str -> {...} testBlock(FunctionLambdaNamesTest:86)]`.

It looks like at some point this mechanism was lost and currently test:
```
  @Test
  public void testBlock() {
    Optional<String> lambdaName = getDescriptionForFunction(str -> {
      if (str.length() > 0) {
        return str.trim();
      } else {
        return str;
      }
    });
    assertTrue(lambdaName.isPresent());
    System.out.println(lambdaName);
  }
```
prints `Optional[trim() testBlock(TestMethodInv:213)]` what is confusing. 

I think we need to bring this mechanism back because this will show up in many real life scenarios.

contrib/parseq-lambda-names/src/test/java/com/linkedin/parseq/lambda/TestMethodRef.java
```diff
@@ -0,0 +1,215 @@
...
+    Optional<String> description = getDescriptionForConsumer(paramMethod(Long.MAX_VALUE, "")::consumer);
+    assertTrue(description.isPresent());
+    assertNameMatch("paramMethod(_,_)::consumer", "testConsumerOnParamMethod", CLASSNAME, description.get().toString());
+  }
+
+  @Test
+  public void testNewInstance() {
+    Optional<String> description = getDescriptionForCallable(new String("abc")::toString);
+    assertTrue(description.isPresent());
+    assertNameMatch("new String()::toString", "testNewInstance", CLASSNAME, description.get().toString());
```
jodzga: Shouldn't this be:
new String(_)::toString

contrib/parseq-lambda-names/src/test/java/com/linkedin/parseq/lambda/TestUnrecognizedLambda.java
```diff
@@ -0,0 +1,110 @@
...
+    assertNameMatch("", "testExpressions", CLASSNAME, description.get().toString());
+  }
+
+  @Test
+  public void testOperations() {
+    MathOperation multiplication = (int a, int b) -> { return a * b; };
+    Optional<String> description = getDescriptionForIntCallable(() -> {return this.operate(5, 3, multiplication);});;
+    assertTrue(description.isPresent());
+    //if operate function's return type is changed to Integer, we aren't able to infer operation
+    assertNameMatch("operate(_,_,_).Integer.valueOf(_)", "testOperations", CLASSNAME, description.get().toString());
```
jodzga: This description looks weird: `operate(_,_,_).Integer.valueOf(_)`
There are two methods `operate(_,_,_)` and `Integer.valueOf(_)` and they are connected with `.`.

Maybe we could detect autoboxing and truncate it from description? I think it would be reasonable to truncate Integer.valueOf(int) if return type of a lambda is Integer. Similar for other primitives e.g. truncate Long.valueOf(long) if return type of a lambda is Long

This way description would be `operate(_,_,_)`.

jodzga:
TupleNTask.java.ftl also needs to be updated e.g.:
```
  /**
   * Equivalent to {@code map("map", f)}.
   * @see #map(String, Function${i})
   */
  default <R> Task<R> map(final Function${i}<<@typeParameters i/>, R> f) {
    return map(tuple -> f.apply(<@tupleArgs i/>));
  }
```
to:
```
  /**
   * Equivalent to {@code map("map", f)}.
   * @see #map(String, Function${i})
   */
  default <R> Task<R> map(final Function${i}<<@typeParameters i/>, R> f) {
    return map("map: " + _taskDescriptor.getDescription(f.getClass().getName()), tuple -> f.apply(<@tupleArgs i/>));
  }
```
so that generated class Tuple2Task has:
```
 /**
   * Equivalent to {@code map("map", f)}.
   * @see #map(String, Function2)
   */
  default <R> Task<R> map(final Function2<T1, T2, R> f) {
    return map("map: " + _taskDescriptor.getDescription(f.getClass().getName()), tuple -> f.apply(tuple._1(), tuple._2()));
  }
```
instead of:
```
  /**
   * Equivalent to {@code map("map", f)}.
   * @see #map(String, Function2)
   */
  default <R> Task<R> map(final Function2<T1, T2, R> f) {
    return map(tuple -> f.apply(tuple._1(), tuple._2()));
  }
```
Otherwise the description comes from the Tuple2Task instead of lambda class passed in as a parameter.

jodzga:
I think in general, if lambda is not a method reference then we should only show last method invocation if it is a on-liner e.g. for:
```
fetchPerson(id).flatMap(person -> createConnectionsSummaries(person.getConnections()));
```
we currently show:
```
map: flatMap: getConnections().createConnectionsSummaries(_) createConnectionsSummaries(Examples:79)
```
but I think it would be better just show:
```
map: flatMap: createConnectionsSummaries(_) createConnectionsSummaries(Examples:79)
```

Also, I know this is not introduced in this PR but it would be great to replace "map: flatMap:" with just "flatMap:".

contrib/parseq-lambda-names/src/test/java/com/linkedin/parseq/lambda/TestMethodRef.java
```diff
@@ -0,0 +1,227 @@
...
+
+  TestMethodRef getTestMethodRef() {
+    return new TestMethodRef();
+  }
+
+  @Test
+  public void testFunctionOnThisChained() {
+    Optional<String> description = getTestMethodRef().getDescriptionForFunction(this::function);
+    assertTrue(description.isPresent());
+    assertNameMatch("function", "testFunctionOnThisChained", CLASSNAME, description.get());
```
jodzga: Can we prefix method references with "::" to more resemble how it looks like in source code e.g. "function" change to "::function"?

src/com/linkedin/parseq/TaskDescriptor.java
```diff
@@ -0,0 +1,23 @@
...
+ * The implementation doesnt need to be thread-safe.
+ *
+ * @author Siddharth Sodhani (ssodhani@linkedin.com)
+ */
+public interface TaskDescriptor {
```
jodzga: I think it would be best to get rid of this interface and replace it with Function<String, String>.
The main disadvantage is that it requires dependency from parseq-lambda-names to parseq main artifact. Currently parseq artifact is included in parseq-lambda-names which is not a good idea because we might have two different versions of parseq main artifact coming from two different class loaders
If we change that and parseq-lambda-names will assume that parseq main artifact is available at runtime we have two other problems. First: what if it isn't? We might see some class loading exceptions. Second: class loading is very tricky and we have to make sure that parseq artifacts are reachable from parseq-lambda-names regardless of which class loaders they are coming from.
Having TaskDescriptor interface is elegant but IMO it is not worth the  additional complexity. If we get rid of this interface the parseq main artifact would be completely independent from parseq-lambda-names

contrib/parseq-lambda-names/src/test/java/com/linkedin/parseq/lambda/TestMethodRef.java
ssodhanilinkedin: Done

src/com/linkedin/parseq/TaskDescriptor.java
ssodhanilinkedin: As discussed, copying TaskDescriptor to contrib parseq-lambda-names project and removing dependency on parseq. Testing in voyager-api.

jodzga:
Ship it!

PR #119, opened 2017-03-13
Optimize task stack trace collection

extesy:
It turns out according to benchmarks that creating temporary exception object is approximately 10x faster than collecting the full stack trace by calling `Thread.getStackTrace()`. The reason is explained in https://stackoverflow.com/questions/26116181/access-stacktraces-with-good-performance - basically, collecting takes only 10% and the rest 90% is spent on converting it into a readable format. Since creating Throwable by itself doesn't do any conversion, it ends up being 10x faster.

### DIFF
--- a/src/com/linkedin/parseq/BaseTask.java
+++ b/src/com/linkedin/parseq/BaseTask.java
@@ -89,7 +89,7 @@ public abstract class BaseTask<T> extends DelegatingPromise<T>implements Task<T>
 
   private volatile TraceBuilder _traceBuilder;
 
-  private final StackTraceElement[] _taskStackTrace;
+  private final Throwable _taskStackTraceHolder;
 
   /**
    * Constructs a base task without a specified name. The name for this task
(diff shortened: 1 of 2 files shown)


src/com/linkedin/parseq/BaseTask.java
```diff
@@ -329,10 +329,12 @@ private void fail(final Throwable error, final TaskLogger taskLogger) {
 
   // Concatenate stack traces if kept the original stack trace from the task creation
   private void appendTaskStackTrace(final Throwable error) {
+    StackTraceElement[] taskStackTrace = this._taskStackTraceHolder != null ? this._taskStackTraceHolder.getStackTrace() : null;
```
jodzga: Very minor:
this._taskStackTraceHolder -> _taskStackTraceHolder

extesy: Fixed!

jodzga:
Can you also add an entry to a CHANGELOG and I'll merge and release it.
Thanks for awesome work!

Issue #140, opened 2017-08-31
Helper method to create a Task instance from a ()->CompletableFuture


jodzga:
This would be a convenience method that would make it easier to integrate ParSeq with other libraries.

mchen07:
Parseq 2.6.30 has addressed this issue.

PR #156, opened 2018-01-23
CompletionStage for Integration

seliang:
Helper method to create a Task instance from a ()->CompletableFuture
#140 opened on Aug 31, 2017 by jodzga


src/com/linkedin/parseq/ParSeqHelper.java
```diff
@@ -0,0 +1,28 @@
...
+import com.linkedin.parseq.promise.Promises;
+import com.linkedin.parseq.promise.SettablePromise;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionStage;
+
+public class ParSeqHelper {
+  /**
+   * Helper method to create a Task instance from a ()->CompletionStage
+   */
+  public static <T> Task<T> toTask(final Callable<CompletionStage<? extends T>> callable)
```
jodzga: I know we discussed it and I suggested to create a helper class but in current form it will be hard to discover this new method.
Could you move this method to Task interface and name it "fromFuture"?
Also please provide two forms: one that accepts Task's description and one that generates it automatically e.g.:
```
public static <T> Task<T> fromFuture(final String desc, final Callable<CompletionStage<? extends T>> callable) {
...
}

public static <T> Task<T> fromFuture(final Callable<CompletionStage<? extends T>> callable) {
  return fromFuture("fromFuture: " + _taskDescriptor.getDescription(callable.getClass().getName()), callable);
}
```

seliang: Got it, this structure makes more sense. I have uploaded the new "fromFuture"

src/com/linkedin/parseq/Task.java
```diff
@@ -1025,6 +1028,43 @@
...
+   * Creates a new task that's value will be set to the value returned
+   * from the CompletionStage.
+   *
+   * Returned task will fail if callable passed in as a parameter throws
+   * an exception.
+   *
+   * @param <T> the type of the return value for this task
+   * @param desc description of the task, it will show up in a trace
+   * @param callable the callable to execute when this task is run
+   * @return the new task that will invoke the callable and complete CompletionStage with its result
```
jodzga: This comment is a bit misleading. It suggests that task will complete CompletionStage. Can you rephrase e.g.:
"the new task that will invoke the callable and complete with result of returned CompletionStage"

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,53 @@ public void onResolved(Promise<String> stringPromise) {
     assertEquals(Boolean.TRUE, resultRef.get());
   }
 
+  @Test
+  public void testToTask() {
```
jodzga: change name of unit tests to reflect new name "fromFuture"

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,53 @@ public void onResolved(Promise<String> stringPromise) {
...
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> {
+        resultRef.set(true);
+        return "";
+      });
+      return completableFuture;
+    });
+    runAndWait("testToTask.success", task);
+    assertEquals(Boolean.TRUE, resultRef.get());
```
jodzga: This test should test that value returned by CompletionStage is exactly the same as the value returned by the task.

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,53 @@ public void onResolved(Promise<String> stringPromise) {
...
+  }
+
+  @Test
+  public void testToTaskWithFailure() {
+    final AtomicReference<Boolean> resultRef = new AtomicReference<Boolean>(false);
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> {
+        try {
+          Thread.sleep(10000);
```
jodzga: In this test CompletionStage should complete with failure. I don't understand why we use Thread.sleep() here.

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,53 @@ public void onResolved(Promise<String> stringPromise) {
...
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> {
+        throw new RuntimeException();
+      });
+      return completableFuture;
+    });
+    runAndWaitException(task, CompletionException.class);
+  }
+
```
jodzga: We also need a test where Callable throws an Exception.

test/com/linkedin/parseq/TestTasks.java
seliang: I want to test whether the task will wait for the future result. I have updated the test

test/com/linkedin/parseq/TestTasks.java
seliang: Add in the to test

test/com/linkedin/parseq/TestTasks.java
seliang: Done

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,59 @@ public void onResolved(Promise<String> stringPromise) {
...
 
+  @Test
+  public void testFromFuture() {
+    String result = "FROMFUTURERESULT";
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> result);
+      return completableFuture;
+    });
+    String taskResult = runAndWait("testToTask.success", task);
```
jodzga: update String name here

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,59 @@ public void onResolved(Promise<String> stringPromise) {
...
+        try {
+          Thread.sleep(1000);
+        } catch (InterruptedException e) {
+          throw new RuntimeException(e);
+        }
+        return result;
+      });
+      return completableFuture;
+    });
+    run(task);
```
jodzga: Please use runAndWait instead.
I understand that you want to check if the task execution took at least 500ms but I think we can remove this part.

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,59 @@ public void onResolved(Promise<String> stringPromise) {
...
+  }
+
+  @Test
+  public void testFromFutureWithTimeConsumingFuture() throws InterruptedException {
+    String result = "FROMFUTURERESULT";
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> {
+        try {
+          Thread.sleep(1000);
```
jodzga: Please decrease it to something smaller e.g. 100ms. We don't want one unit test to take 1s.

seliang: Fixed

src/com/linkedin/parseq/Task.java
```diff
@@ -49,6 +50,8 @@
 import com.linkedin.parseq.trace.Trace;
 import com.linkedin.parseq.trace.TraceBuilder;
 
+import static com.linkedin.parseq.Task.*;
```
jodzga: Let's not do wildcard imports.

test/com/linkedin/parseq/TestTasks.java
```diff
@@ -89,6 +91,57 @@ public void onResolved(Promise<String> stringPromise) {
...
 
+  @Test
+  public void testFromFuture() {
+    String result = "FROMFUTURERESULT";
+    Task<String> task = Task.fromFuture(() ->{
+      CompletableFuture<String> completableFuture
+          = CompletableFuture.supplyAsync(() -> result);
+      return completableFuture;
+    });
+    runAndWait("testToTask.success", task);
```
jodzga: Replace "testToTask.success" with "testFromFuture"

seliang: Fixed，misunderstood your comment at first

jodzga:
We should also add reverse conversion: from Promise to CompletableFuture by adding default method toFuture() to Promise interface.

xma000:
I suppose the implementation is alone the line of

public CompletableFuture<T> toFuture() {
  final CompletableFuture<T> future = new CompletableFuture<>();
  addListener(p -> {
    if (!p.isFailed()) {
      future.complete(p.get());
    }
    else {
      future.completeExceptionally(p.getError());
    }
  });
  return future;
}

mchen07:
That is a neat suggestion. Implementing CompletionStage interface in ParSeq seems quite challenging. This seems the right direction to go.

jodzga:
Yes, this is exactly what I meant. This should be a default method on a Promise interface.

mchen07:
For API symmetry, I may prefer `toFuture` to be symmetric to `fromFuture`. `toCompletionStage` is not too friendly for API consumer.

xma000:
I just noticed `fromFuture` uses `CompletionStage` in its signature.

See: https://github.com/linkedin/parseq/commit/61bed2e

mchen07:
Yes, from application side that is frequently using CompletableFuture, they can directly use `fromFuture` without any conversion. But to use `toCompletionStage`, they need to explicitly convert to CompletableFuture. If we decide to go with CompetionStage route, then we need to change names in both places to be consistent.

seliang:
I have convert the code from CompletionFuture to CompletionStage.
Method names are fromCompletionStage and toCompletionStage

src/com/linkedin/parseq/Task.java
```diff
@@ -1025,6 +1026,44 @@
...
+        else {
+          promise.done(value);
+        }
+      });
+      return promise;
+    });
+  }
+
+  /**
+   * Equivalent to {@code fromCompletionStage("fromFuture", callable)}.
```
mchen07: Fix comments. "fromFuture" to "fromCompletionStage"

PR #160, opened 2018-02-27
Added support for sub-resources in ParSeqRestClient configuration.

jodzga:
With this change user can now use `:` character to separate parent resource from it's children.

For example, configuration key: `*.*/assets:media.GET` will apply to all outgoing GET requests of `media` sub-resource of `assets` resource.
Similarly, configuration key: `*.*/assets:meta.GET` will apply to all outgoing GET requests of `meta` sub-resource of `assets` resource.
Finally, configuration specified on a parent resource will apply to all it's children, e.g. configuration key: `*.*/assets.GET` will apply to all outgoing GET requests of both `media` and `meta` sub-resources of an `assets` resource. This behavior is backwards compatible.


contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/config/RequestConfigTree.java
```diff
@@ -94,31 +95,35 @@ void add(RequestConfigElement element) {
...
+  /**
+   * This method recursively uses given resolver to resolve a config by given name taking into account
+   * syntax of sub-resource names. For example, for given name: Optional.of("foo:bar:baz") it will make
+   * the following resolver calls:
+   * - resolver(Optional.of("foo:bar:baz"))
+   * - resolver(Optional.of("foo:bar"))
+   * - resolver(Optional.of("foo"))
+   * - resolver(Optional.empty())
+   */
+  <X> Optional<X> resolveRecursivelyName(Optional<String> name, Function<Optional<String>, Optional<X>> resolver) {
```
mchen07: Why do we need generics X here? Is this always `Optional<ConfigValue<T>>`?

mchen07: `resolveNameRecursively` sounds more natural.

contrib/parseq-restli-client/README.md
```diff
@@ -23,32 +23,39 @@ Each property is defined by a set of Key-Value pairs where Key has the following
...
 
 ```
 grammar RequestConfigKey;
 
-key             : inbound '/' outbound EOF;
-inbound         : ( Name | '*' ) '.' ( operationIn | '*' );
-outbound        : ( Name | '*' ) '.' ( operationOut | '*' );
-operationIn     : simpleOp | complex | httpExtraOp;
+key 			: inbound '/' outbound EOF;
+inbound			: ( resource | '*' ) '.' ( operationIn | '*' );
```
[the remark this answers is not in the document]
jodzga: I use local markdown renderer to check how rendered page looks like and it seems to be OK. If after checking in the lines are not aligned I'll fix them then.

contrib/parseq-restli-client/src/main/java/RequestConfigKey.tokens
```diff
@@ -21,27 +21,29 @@ T__19=20
 T__20=21
```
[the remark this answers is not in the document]
jodzga: There are pros and cons of having generated code checked in. For example, it is sometimes useful to see in history how those files have changed. Also, it is more intuitive to be able to find source of every file on github vs having to generate some source files locally.
I would leave it as it is for now because it seems to be orthogonal to this PR and maybe create an issue to reconsider in the future if generated source files should be checked in?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/config/RequestConfigTree.java
jodzga: For the logic of this method the only relevant aspect of return type is that it is Optional, thus it is defined as Optional<X>.
I realize that it might be easier to read in the context of surrounding code if it was Optional<ConfigValue<T>>, so I changed it to Optional<ConfigValue<T>>.

contrib/parseq-restli-client/src/main/antlr4/com/linkedin/restli/client/config/RequestConfigKey.g4
```diff
@@ -1,8 +1,9 @@
 grammar RequestConfigKey;
 
 key 			: inbound '/' outbound EOF;
-inbound			: ( Name | '*' ) '.' ( operationIn | '*' );
-outbound		: ( Name | '*' ) '.' ( operationOut | '*' );
+inbound			: ( resource | '*' ) '.' ( operationIn | '*' );
```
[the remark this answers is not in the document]
jodzga: Changed to restResource.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/config/RequestConfigTree.java
jodzga: Will rename.

contrib/parseq-restli-client/src/main/java/RequestConfigKey.tokens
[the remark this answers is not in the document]
mchen07: Sure. I didn't mean to address this in this pull request. Just would like to have consistent convention to follow.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/config/RequestConfigTree.java
mchen07: Thanks.

jodzga:
Regarding: FAILED: testStallsWithStuckThread
This is a flaky unit test. It does depend on wall clock which means it is not reliable in virtual environment where process can be CPU deprived.
I'll create a ticket to fix this  test.

Issue #144, opened 2017-10-16
Increase arity of Task.par()


jodzga:
Currently Task.par() takes up to 9 parameters.
Users requested need for up to 11 parameters in real life.

mchen07:
ParSeq 2.6.29 has addressed this issue.

PR #168, opened 2018-03-21
increase arity of Task.par() from 9 to 11 #144

bolobobo:
Currently Task.par() takes up to 9 parameters and this new task is to generate up to Par11Task and Tuple11Task variants.

(1) change the parameter of the arity in fmpp, generate new class ParNTask and TupleNTask automatically.
(2) add similar code like 'par' method in Task.java to extend to support 10 and 11 parameter which is type-safety.
(3) add test about new 'par' method, like testPar10AndThen, testPar11, etc.


fmpp/data/fmpp.tdd
```diff
@@ -1 +1 @@
-max: 9 <#-- must be at least 3 -->
+max: 11 <#-- must be at least 3 -->
```
mchen07: Can we bump this to 15? That should be our maximum threshold.

contrib/parseq-examples/src/main/java/com/linkedin/parseq/example/jguanexample/IntroCase.java
```diff
@@ -0,0 +1,10 @@
+package com.linkedin.parseq.example.jguanexample;
+
+import com.linkedin.parseq.Engine;
+import com.linkedin.parseq.Task;
+import com.linkedin.parseq.example.common.AbstractExample;
+import com.linkedin.parseq.example.common.ExampleUtil;
+
+
+public class IntroCase {
```
mchen07: Is this some test file?

mchen07:
Also, based on guideline, please update CHANGELOG.md file about this change.

PR #172, opened 2018-04-07
PlanCompletionListener should be invoked when batch task is finished.

mchen07:


contrib/parseq-batching/src/main/java/com/linkedin/parseq/batching/BatchingStrategy.java
```diff
@@ -1,16 +1,5 @@
 package com.linkedin.parseq.batching;
 
-import java.util.ArrayList;
```
jodzga: It would be great to avoid stylistic changes because it is hard to figure out what has really changed and what is just reformatting.

mchen07: Wait a second, in my local, testLongRunningBatchTaskSuccess should fail without my change, is it passing for you? Then something is wrong!

mchen07: Ignore my previous comment, that is right, testLongRunningBatchTaskSuccess works fine even with previous code since there is no task cancellation and ParTask will wait for the second task promise complete. testLongRunningBatchTaskFailure is the original bug where task cancellation makes ParTask promise complete quickly.

PR #173, opened 2018-04-11
fix tracevis name truncation

bolobobo:
According to the previous design of string truncation,  the length of the name string is less than 255 characters. 
Set the width of grid's right margin to a reasonable value instead of 250.


bolobobo:
Since each character's width is 7~8px,  set the whole width of the longest string(255 chars * 8 + buffer) to 2130.

mchen07:
Please put this into comments in the js code.

PR #176, opened 2018-04-25
Use D2 per-request timeout if enabled.

mchen07:
This change is [Reviewable](https://reviewable.io/reviews/linkedin/parseq/176)


mchen07:
I initially thought that we would like to first turn on for some apps, if we would like to switch globally, it makes sense to put it in global config.

jodzga:
Hmm, the commenting feature on Github is weird - I don't know how to respond to you comment to form a thread.

Anyway, the rollout will certainly be per app. By global I mean global inside single JVM. Currently we can specify this configuration per specific endpoint / operation.

mchen07:
I also don't like PR comment feature, not sure how to reply to a specific comment. I got your point, have changed to use global configuration instead of fine-grained configuration per instance.

davidhoa:
Hmm, I thought that they improved it. Try reviewable.io<http://reviewable.io>,
I’ve been told it’s better for github reviews. (Disclaimer: my neighbor runs that company).

David

On May 11, 2018, at 6:38 PM, Min Chen <notifications@github.com<mailto:notifications@github.com>> wrote:


I also don't like PR comment feature, not sure how to reply to a specific comment. I got your point, have changed to use global configuration instead of fine-grained configuration per instance.

hiteshsharma:
Does this change mean that if the upstream is using a parSeq timeout config for a Rest.li call then the Parseq task run by downstream service will respect that timeout and terminate and return?

mchen07:
This is still at client side, not server. In your example, upstream callback may get invoked earlier if you have specified a shorter timeout than that defined by its downstream service, but downstream service will continue to process. In our roadmap, we do have another feature to propagate a SLA to downstream services so that downstream service can determine to terminate early or not run at all based on propagated SLA and its resource budget. These two features are related, but not the same.

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/linkedin/parseq/176#-:-LCUXi8oCH_kR0jeesDm:bhr1llf)*

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -197,10 +207,49 @@ static String generateTaskName(final Request<?> request) {
...
+      requestTask = createBatchableTask(name, request, requestContext, config);
+    } else {
+      requestTask = Task.async(name, () -> sendRequest(request, requestContext));
+    }
+    if (!ParSeqGlobalConfiguration.isD2RequestTimeoutEnabled()) {
+      return withTimeout(requestTask, timeout);
     } else {
-      return withTimeout(Task.async(name, () -> sendRequest(request, requestContext)), config);
+      // transform TimeoutException to be compatible with current Task timeout exception message
+      return requestTask.transform((Try<Response<T>> tryGet) -> {
```
jodzga: We should get this transform Task a name - preferably the same as the one that original withTimeout had i.e. so that it includes the configuration key that caused timeout to be applied.

contrib/parseq-restli-client/src/test/java/com/linkedin/restli/client/CapturingRestClient.java
```diff
@@ -12,6 +12,10 @@
 import com.linkedin.restli.client.multiplexer.MultiplexedRequest;
 import com.linkedin.restli.client.multiplexer.MultiplexedResponse;
 
+
+/**
+ * The type Capturing rest client.
```
jodzga: Hmmm, "type Capturing"? Can you expand this javadoc comment?

contrib/parseq-restli-client/src/test/java/com/linkedin/restli/client/ParSeqRestClientBatchingIntegrationTest.java
```diff
@@ -55,7 +55,7 @@ public void testGetSubResourceRequests() {
     Tuple2Task<Response<Message>,Response<Message>> task = Task.par(associationsGet("a", "b", "x"), associationsGet("a", "b", "y"));
     if (expectBatching()) {
       runAndWaitException(task, RestLiResponseException.class);
-      assertTrue(task.getError().getCause().getMessage().contains("associationsSub?ids=List(x,y)"));
```
jodzga: Why is this change needed? Does it mean that this change is not backwards compatible? For example some users might be calling task.getError().getCause().getMessage().

src/com/linkedin/parseq/ParSeqGlobalConfiguration.java
```diff
@@ -86,4 +88,36 @@ public static void setTrampolineEnabled(boolean enabled) {
...
+   * Once enabled, the timeout used in Parseq can be passed down into rest.li R2D2 layer
+   * and free up memory sooner than currently possible. For example, if the server defined
+   * requestTimeout is 10s, but the client only wants to wait 250ms, parseq rest client
+   * could tell r2d2 to trigger the user callback after 250 ms instead of 10s, which would
+   * help us handle the case of slow downstream services, which today cause memory problems
+   * for high qps upstream services because more objects are being held longer in memory.
+   *
+   * @return true if d2 per-request timeout is currently enabled
+   */
+  public static boolean isD2RequestTimeoutEnabled() {
```
jodzga: I think this should be moved to partseq-rest-client contrib project.

contrib/parseq-restli-client/src/test/java/com/linkedin/restli/client/ParSeqRestClientBatchingIntegrationTest.java
mchen07: This change is required due to rest.li commit https://github.com/linkedin/rest.li/commit/7498b2196152f64bf36b3f67683673c26228cbdb. The RestLiResponseException.getCause().getMessage will not contain query parameters. This backward incompatibility comes from pegasus (indicated by its major version bump). If the users have bumped to pegasus 19.* or above and their existing code relies on such specific message, they will have to fix this.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -65,17 +73,19 @@
 
   private final Client _client;
   private final BatchingMetrics _batchingMetrics = new BatchingMetrics();
-  private final RequestConfigProvider _clientConfig;
+  private final RequestConfigProvider _requestConfigProvider;
+  private boolean _d2RequestTimeoutEnabled = false;
```
jodzga: should be final

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +205,60 @@ static String generateTaskName(final Request<?> request) {
...
+      String desc = timeout.getSource().map(src -> " src: " + src).orElse("");
+      String timeoutTaskName = "withTimeout " + Math.min(timeout.getValue(), Integer.MAX_VALUE)
+          + TimeUnitHelper.toString(TimeUnit.MILLISECONDS) + desc;
+      return task.transform(timeoutTaskName, (Try<Response<T>> tryGet) -> {
+        if (tryGet.isFailed()) {
+          if (tryGet.getError() instanceof TimeoutException) {
+            String timeoutExceptionMessage = "task: '" + task.getName() + "' " + timeoutTaskName;
+            return Failure.of(Exceptions.timeoutException(timeoutExceptionMessage));
+          } else {
+            return Failure.of(tryGet.getError());
```
jodzga: Can we just return tryGet value here?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +205,60 @@ static String generateTaskName(final Request<?> request) {
...
+          }
+        } else {
+          return Success.of(tryGet.get());
```
jodzga: Can we just return tryGet value here?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +205,60 @@ static String generateTaskName(final Request<?> request) {
...
+    Number contextTimeout = (Number) requestContext.getLocalAttr(R2Constants.REQUEST_TIMEOUT);
+    int timeoutBound = (contextTimeout != null) ? contextTimeout.intValue() : Integer.MAX_VALUE;
+    // d2 request timeout can only accept Integer timeout
+    requestContext.putLocalAttr(R2Constants.REQUEST_TIMEOUT, Math.min(timeout, timeoutBound));
+  }
+
+  private <T> Task<Response<T>> createTaskWithTimeout(final String name, final Request<T> request,
+      final RequestContext requestContext, RequestConfig config) {
+    ConfigValue<Long> timeout = config.getTimeoutMs();
+    if (timeout.getValue() != null && timeout.getValue() > 0 && _d2RequestTimeoutEnabled) {
```
jodzga: This if statement can be removed and reconcileRequestTimeout method invocation can be moved inside withD2Timeout inside "if (timeout.getValue() != null && timeout.getValue() > 0)" condition.

mchen07: hmm... reconcileRequest needs to update requestContext, which is used to create the parseq Task passed to withD2Timeout. Thus it will be too late to call reconcileRequestTimeout inside withD2Timeout?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Fixed

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: True. Combined two if-else into one.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Fixed.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
jodzga: Yup, you are right.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +207,56 @@ static String generateTaskName(final Request<?> request) {
     }
   }
 
+  private <T> Task<Response<T>> withD2Timeout(final Task<Response<T>> task, ConfigValue<Long> timeout) {
+    if (timeout.getValue() != null && timeout.getValue() > 0) {
+      // transform TimeoutException to be compatible with current Task timeout exception message
+      String desc = timeout.getSource().map(src -> " src: " + src).orElse("");
+      String timeoutTaskName = "withTimeout " + Math.min(timeout.getValue(), Integer.MAX_VALUE)
```
jodzga: What is a purpose of Math.min()? Is D2 timeout expect Integer type?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +207,56 @@ static String generateTaskName(final Request<?> request) {
...
+
   private <T> Task<Response<T>> createTask(final String name, final Request<T> request,
       final RequestContext requestContext, RequestConfig config) {
     LOGGER.debug("createTask, name: '{}', config: {}", name, config);
+    return createTaskWithTimeout(name, request, requestContext, config);
+  }
+
+  // reconcile timeout specified in requestContext with the given timeout
+  private void reconcileRequestTimeout(final RequestContext requestContext, Long timeout) {
+    Number contextTimeout = (Number) requestContext.getLocalAttr(R2Constants.REQUEST_TIMEOUT);
```
jodzga: I think we need to be defensive and catch ClassCastException.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +207,56 @@ static String generateTaskName(final Request<?> request) {
...
+    int timeoutBound = (contextTimeout != null) ? contextTimeout.intValue() : Integer.MAX_VALUE;
+    // d2 request timeout can only accept Integer timeout
+    requestContext.putLocalAttr(R2Constants.REQUEST_TIMEOUT, Math.min(timeout, timeoutBound));
```
jodzga: Does this have the right type? Should it be integer?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -194,13 +207,56 @@ static String generateTaskName(final Request<?> request) {
...
   private <T> Task<Response<T>> createTask(final String name, final Request<T> request,
       final RequestContext requestContext, RequestConfig config) {
     LOGGER.debug("createTask, name: '{}', config: {}", name, config);
+    return createTaskWithTimeout(name, request, requestContext, config);
+  }
+
+  // reconcile timeout specified in requestContext with the given timeout
+  private void reconcileRequestTimeout(final RequestContext requestContext, Long timeout) {
+    Number contextTimeout = (Number) requestContext.getLocalAttr(R2Constants.REQUEST_TIMEOUT);
+    int timeoutBound = (contextTimeout != null) ? contextTimeout.intValue() : Integer.MAX_VALUE;
```
jodzga: I think it could be simplified. Do we need to check with Integer.MAX_VALUE?
Isn't Math.min(timeout, contextTimeout) enough (except that it has to be properly cast to a common type).

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
jodzga: The task name does not take into consideration reconciled value from RequestContext. Ideally we could include reconciled value and D2 as source if it was the source of a timeout.

mchen07: Yes. In rest.li code, we have made assumption that request timeout should be specified as Integer, such as https://github.com/linkedin/rest.li/blob/master/r2-netty/src/main/java/com/linkedin/r2/transport/http/client/HttpClientFactory.java#L1067, https://github.com/linkedin/rest.li/blob/master/d2/src/main/java/com/linkedin/d2/balancer/clients/RequestTimeoutClient.java#L143, where we directly invoke Number.intValue to get the timeout set in RequestContext. If we don't use Math.min() here, we will get Overflow in RequestTimeoutClient D2 code.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Agree. Will guard this.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Java autoboxing allows us to directly put primitive int into a map?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: what happens if requestContext does not set R2Constants.REQUEST_TIMEOUT? In those case, we should  use the configured timeout.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Make sense. Let me enhance this.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Let me make this explicit as Math.min(int, int), same for other places where Math.min is used.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Simplified the code to make this easier to read.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -181,26 +195,72 @@ static String generateTaskName(final Request<?> request) {
...
+      LOGGER.error("Ignore invalid value for REQUEST_TIMEOUT set in request context: " + requestTimeout);
+    }
+    requestContext.putLocalAttr(R2Constants.REQUEST_TIMEOUT, d2Timeout);
+    return d2Timeout;
+  }
+
+  private <T> Task<Response<T>> createTaskWithTimeout(final String name, final Request<T> request,
+      final RequestContext requestContext, RequestConfig config) {
+    ConfigValue<Long> timeout = config.getTimeoutMs();
+    int reconciledTimeout = 0;
```
jodzga: I think we need to be defensive and guard against timeout in request context <= 0 and ignore it (and log) if that is the case.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -181,26 +195,72 @@ static String generateTaskName(final Request<?> request) {
...
+    }
+    if (timeout.getValue() == null || timeout.getValue() <= 0) {
+      // no timeout configured
+      return requestTask;
     } else {
-      return withTimeout(Task.async(name, () -> sendRequest(request, requestContext)), config);
+      if (!_d2RequestTimeoutEnabled) {
+        return withTimeout(requestTask, timeout);
+      } else {
+        assert reconciledTimeout > 0;
```
jodzga: I think it is possible that reconciledTimeout = 0 if timeout in request context was 0?

PR #177, opened 2018-04-26
add version number to tracevis server jar

bolobobo:
Currently we build the parseq-tracevis-server-jar-with-dependencies.jar without version number. 
To make it easier to be managed and deployed, give version number to this jar.

### DIFF
--- a/contrib/parseq-tracevis-server/pom.xml
+++ b/contrib/parseq-tracevis-server/pom.xml
@@ -101,7 +101,6 @@
 					<descriptors>
 						<descriptor>assembly-descriptor.xml</descriptor>
 					</descriptors>
-					<finalName>parseq-tracevis-server</finalName>
 					<archive>
 						<manifest>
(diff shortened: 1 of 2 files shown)


contrib/parseq-tracevis-server/pom.xml
```diff
@@ -101,7 +101,6 @@
 					<descriptors>
 						<descriptor>assembly-descriptor.xml</descriptor>
 					</descriptors>
-					<finalName>parseq-tracevis-server</finalName>
```
mchen07: Interesting, so where is this "*-with-dependencies.jar" name defined?

bolobobo: There is another file named "assembly-descriptor.xml" which defines the dependencies.jar.

bolobobo:
I tested it locally. It did generate all the .jar files in my local maven repo. In my understanding, if it work locally, it should work remotely.

PR #185, opened 2018-05-15
Support multi-colo service call in request config.

mchen07:


contrib/parseq-restli-client/src/main/antlr4/com/linkedin/restli/client/config/RequestConfigKey.g4
```diff
@@ -3,7 +3,7 @@ grammar RequestConfigKey;
 key 			: inbound '/' outbound EOF;
 inbound			: ( restResource | '*' ) '.' ( operationIn | '*' );
 outbound		: ( restResource | '*' ) '.' ( operationOut | '*' );
-restResource 	: Name ( ':' Name )*;
+restResource 	: Name ( '-' Name )* ( ':' Name ( '-' Name )* )*;
```
jodzga: I think colo can be only specified on resource name, not sub-resource, so it should be:
restResource 	: Name ( '-' Name )* ( ':' Name )*;

mchen07: Yes, you are right. sub-resource should not have multi-colo support, fixed.

mchen07:
Service deployed locally may not be enough to test this feature since we need a service that has done multi-colo announcing. I will see if I can send a d2 request against a real multi-colo service to make sure that the integration flow works as expected.

PR #186, opened 2018-05-18
Prevent cross-plan task sharing if configured.

mchen07:


src/com/linkedin/parseq/BaseTask.java
```diff
@@ -389,18 +391,24 @@ private void appendTaskStackTrace(final Throwable error) {
...
-  protected boolean transitionRun(final TraceBuilder traceBuilder) {
+  protected boolean transitionRun(final Context context) {
     State state;
     State newState;
     do {
       state = _stateRef.get();
       if (state.getType() != StateType.INIT) {
+        // prevent cross-plan task sharing if enabled
+        if (!ParSeqGlobalConfiguration.isAllowCrossPlanSharingEnabled() &&
+            _startPlanId != null && !_startPlanId.equals(context.getPlanId())) {
```
jodzga: This is not bulletproof because it might be the case that _startPlanId == null here.
Since State is CASed atomically the _startPlanId should belong to State class. Then it would be ensured that _startPlanId != null.

src/com/linkedin/parseq/BaseTask.java
```diff
@@ -389,18 +391,24 @@ private void appendTaskStackTrace(final Throwable error) {
...
+          throw new IllegalStateException(this.toString() + " should not be shared across plans!!");
```
jodzga: We can't just throw exception here because it might leave Task in inconsistent state. We also need to log error when this problem occurs. It would be best to change exception to be a checked exception and wrap all calls to this method with try-catch to make sure Task's state is recovered.

src/com/linkedin/parseq/ParSeqGlobalConfiguration.java
```diff
@@ -26,6 +26,8 @@
 public final class ParSeqGlobalConfiguration {
   private static volatile boolean _crossThreadStackTracesEnabled = false;
   private static volatile boolean _trampolineEnabled = false;
+  // default set to true to keep backward compatibility
+  private static volatile boolean _allowCrossPlanSharing = true;
```
jodzga: _allowCrossPlanSharing -> _allowCrossPlanTaskSharing

src/com/linkedin/parseq/BaseTask.java
```diff
@@ -18,6 +18,7 @@
 
 import java.util.Collection;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
```
jodzga: Not needed?

src/com/linkedin/parseq/BaseTask.java
mchen07: That was my first attempt to put _startPlanId in State class, but then I realized that all those static final State instances  (INIT, RUN, DONE we used broadly in transition) are not quite useful anymore. Not sure about impact of removing all those static final State instance.

src/com/linkedin/parseq/BaseTask.java
mchen07: hmm... that is leftover from my earlier attempt of adding _planId in State class.

src/com/linkedin/parseq/BaseTask.java
mchen07: Based on our discussion, I have put startPlanId as part of State class, removed all those PENDING, DONE, RUNNING static final State instance with only INIT left.  This will make State and startPlanId transition atomic. Thanks for pointing this out.

src/com/linkedin/parseq/BaseTask.java
mchen07: Added a checked exception class CrossPlanTaskSharingException.

mchen07:
@jodzga Your observation is correct, after adding some testcases to cover map, recover, etc parseq functionalities, I realized a fundamental flaw of current recovery mechanism after throwing an exception when we detected such cross-plan task sharing usage, that is, Promise related to the shared task can only be completed once, and when we detected this incorrect usage, all the PromiseListeners associated with this shared task have been properly set up. It will be quite challenging to notify some listeners about this exception and at the same time guarantee that all tasks are not left in a dangling state. For FusionTask where we are propagating results in a more efficient way, we may be able to work around this a bit, as I have attempted in the recent change and manifested in the new test added. But for async task created by Task.async, flatten, recoverWith, etc, it is not easy to repair this to make one plan success and the conflicting plan fail and meanwhile maintain task state integrity as well as task trace consistency. Based on this observation, I am wondering maybe we need to change the strategy how we want to address this issue, maybe doing some dramatic logging (like repeated logging using some ScheduledExecutorService) to get attention from incorrect users is the best we can do here with minimal side effects.

nickibi:
We are closing this old PR, @mchen07 if you have any concerns feel free to reopen it.

PR #195, opened 2018-09-11
Update parseq d2 timeout by setting REQUEST_TIMEOUT_IGNORE_IF_HIGHER flag properly.

mchen07:
This PR includes two major fixes:

1. Bump to pegasus 24.0.2, which introduces a new d2 per-request timeout flag REQUEST_TIMEOUT_IGNORE_IF_HIGHER to indicate whether we would like to pass down parseq timeout down to R2D2 layer or not.
2. Remove all those generated data template and request builder classes in src/test folder, since now pegasus restli-int-test-api artifact (with classifier "all") is self-contained with those classes to avoid complication in bumping pegasus version in parseq.


contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -220,44 +220,51 @@ static String generateTaskName(final Request<?> request) {
...
     }
     Task<Response<T>> requestTask;
     if (RequestGroup.isBatchable(request, config)) {
       requestTask = createBatchableTask(name, request, requestContext, config);
     } else {
       requestTask = Task.async(name, () -> sendRequest(request, requestContext));
     }
-    if (timeout.getValue() == null || timeout.getValue() <= 0) {
-      // no timeout configured
+    if (timeout.getValue() == null || timeout.getValue() <= 0 || hasPerRequestTimeout) {
```
alexjinghn: If request comes with a timeout in requestContext and parseq timeout is also configured here, then the timeout that comes with the request would win. I don't have enough context but does this feel very natural?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -220,44 +220,51 @@ static String generateTaskName(final Request<?> request) {
...
+      // no timeout configured or per-request timeout already specified in request context
       return requestTask;
     } else {
       if (!_d2RequestTimeoutEnabled) {
```
alexjinghn: Seems that if `d2RequestTimeout` is not enabled, it will just take out the timeout value and call `withTimeout`. This function looks really messy because of the need to put the right timeout in the requestContext. Can we toggle on `d2RequestTimeoutEnabled` and have two separate functions to improve cleanness?

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: semantically I think that this behavior makes sense. Per-request configuration always takes precedence over general request pattern configuration.

contrib/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
mchen07: Thanks for suggestion, I will split them into two functions to make it cleaner.

PR #197, opened 2018-09-19
Update README file and fix vulnerable dependencies in package.json

bolobobo:
Since parseq is migrated to Gradle, some link path in Readme file need to be updated. 
Also, parseq-tracevis has vulnerable dependencies issue, so upgrade the packages.


mchen07:
LGTM.

PR #199, opened 2018-09-27
Support plan based rate limiting.

mchen07:
This PR supports max concurrency plan configured based on plan class.

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/internal/DefaultPlanBasedRateLimiter.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/internal/DefaultPlanBasedRateLimiter.java
@@ -1,0 +1,49 @@
+package com.linkedin.parseq.internal;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+
+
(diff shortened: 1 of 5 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/internal/DefaultPlanBasedRateLimiter.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public boolean tryAcquire(String planClass) {
+    return _concurrentPlansPerClass.containsKey(planClass) ?
+        _concurrentPlansPerClass.get(planClass).tryAcquire() : true;
+  }
+
+  @Override
+  public boolean tryAcquire(String planClass, long timeout, TimeUnit unit) throws InterruptedException {
+    return _concurrentPlansPerClass.containsKey(planClass) ?
```
seliang: When tryAcquire a class which is not contains in the map, we should return true? Do we need to insert that class into the map?

mchen07: Yes, we need to return true, because that means that user does not specify concurrency limit to that plan class. We cannot add it to the map since that map should reflect what user configured for parseq Engine.

PR #200, opened 2018-09-29
Configure unit test in Gradle

bolobobo:
Set up testNG in build.gradle to do unit test and change the test configuration in parseq-lambda-names module.

### DIFF
--- a/subprojects/parseq/build.gradle
+++ b/subprojects/parseq/build.gradle
@@ -44,10 +44,6 @@ dependencies {
   testCompile group: "org.slf4j", name: "slf4j-log4j12", version: "1.7.25"
 }
 
-test {
-  useTestNG()
-}
-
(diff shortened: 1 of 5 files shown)


subprojects/parseq/build.gradle
```diff
@@ -65,9 +61,6 @@ sourceSets {
   test {
       compileClasspath += sourceSets.testUtils.output
       runtimeClasspath += sourceSets.testUtils.output
-      java {
-        srcDirs = ['src/test-utils/java']
```
mchen07: What is this change for?

bolobobo: That's the test utility which should not be included into "test" source code. "test" and "test-utility" should be in separate directory. This is a previous configuration mistake.

Issue #133, opened 2017-08-08
Investigate usage of Task.value(null)


jodzga:
Task.value(null) shows up in real code in places where there are deficiencies in ParSeq's API, see #132 or where it is not clear which existing API is supposed to be used.
This work is to investigate usage of Task.value(null) and if possible improve ParSeq API to avoid it's usage.

Issue #132, opened 2017-08-08
Add Task.sideEffect()


jodzga:
Existing API to create a side effect is Task.withSideEffect(). It requires an existing Task instance to begin with what leads to unnatural constructs such as: Task.value(null).withSideEffect().
ParSeq should provide an API for creating side effect in situations where there is no existing Task instance.

saponniah:
Can the usecase for using `Task.value(null).withSideEffect()` be achieved by `Task.action()`? How will the new `Task.sideEffect()` be different? The only difference I see is that the `Task<Void>` returned from `action()` will be marked as failed during a failure when the current side effect task's completion status is never exposed.

jodzga:
The Task.withSideEffect() have behavior that is very different than "non side effect" task: http://linkedin.github.io/parseq/javadoc/latest/com/linkedin/parseq/Task.html#withSideEffect-java.lang.String-com.linkedin.parseq.function.Function1-

junchuanwang:
Close this as already implemented in #204

PR #204, opened 2019-01-27
Add support for static withSideEffect

hiteshsharma:
For #132 and #133 

Added support for  `static` `Task.withSideeffect` to replace using Parseq API for cases where other usages like `Task.value(null).withSideEffect(...)` were being used.

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -338,9 +338,9 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * @param func function to be applied to get side effect task
    * @return a new Task that will run the side effect Task
    */
-  static <T> Task<T> withSideEffect(final String desc, final Callable<Task<T>> func) {
+  static <T> Task<Void> withSideEffect(final String desc, final Callable<Task<T>> func) {
     ArgumentUtil.requireNotNull(func, "function");
-    final Task<T> sideEffectWrapper = async(desc, ctx -> {
+    final Task<Void> sideEffectWrapper = async(desc, ctx -> {
       Task<?> sideEffect = func.call();
       ctx.runSideEffect(sideEffect);
 
@@ -355,7 +355,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * Equivalent to {@code Task.withSideEffect("withSideEffect", func)}.
    * @see Task#withSideEffect(String, Callable)
    */
-  static <T> Task<T> withSideEffect(final Callable<Task<T>> func) {
+  static <T> Task<Void> withSideEffect(final Callable<Task<T>> func) {
     return Task.withSideEffect("withSideEffect", func);
   }
 
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -316,6 +316,50 @@ public interface Task<T> extends Promise<T>, Cancellable {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -315,6 +315,50 @@
...
+  static <T> Task<T> withSideEffect(final String desc, final Callable<Task<T>> func) {
+    ArgumentUtil.requireNotNull(func, "function");
+    final Task<T> sideEffectWrapper = async(desc, ctx -> {
+      Task<?> sideEffect = func.call();
+      ctx.runSideEffect(sideEffect);
+
+      return Promises.value(null);
+    });
+
+    sideEffectWrapper.getShallowTraceBuilder().setTaskType(TaskType.WITH_SIDE_EFFECT.getName());
```
mchen07: indent

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -315,6 +315,50 @@
...
+    return sideEffectWrapper;
+  }
+
+  /**
+   * Equivalent to {@code Task.withSideEffect("sideEffect", func)}.
+   * @see Task#withSideEffect(String, Callable)
+   */
+  static <T> Task<T> withSideEffect(final Callable<Task<T>> func) {
+    return Task.withSideEffect("sideEffect", func);
```
mchen07: Should we use "withSideEffect" as description to be consistent with current .withSideEffect?

hiteshsharma: fixed

hiteshsharma:
@mchen07, comments addressed. Please take a look when possible

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -315,6 +315,50 @@
...
+   * and the Task can execute in background. In this case, caller should not be concerned with success or failure of
+   * background Task.
+   *
+   * Static side effect can be thought of as a side effect being attached to an empty Task. like:
+   * {@code Task.value(null).withSideEffect(() -> Task.value("hello world")); }
+   * @param desc description of a side effect, it will show up in a trace
+   * @param func function to be applied to get side effect task
+   * @return a new Task that will run the side effect Task
+   */
+  static <T> Task<T> withSideEffect(final String desc, final Callable<Task<T>> func) {
```
mchen07: Should this return a `Task<Void>` instead? This task value should not depend on side effect task.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
mchen07: Maybe the github diff display issue.

subprojects/parseq/src/test/java/com/linkedin/parseq/AbstractTaskTest.java
```diff
@@ -297,6 +297,29 @@ public void testWithSideEffectFailure(int expectedNumberOfTasks) throws Exceptio
...
+
+    run(sideEffect);
+    assertFalse(sideEffect.isDone());
+    sideEffect.await();
+    assertTrue(sideEffect.isDone());
+    assertNull(sideEffect.get());
+  }
+
+  @Test
+  public void testStaticSideEffectFailureIsIgnored() throws Exception {
```
mchen07: Maybe  also add a test in `TestPlanCompletionListener` to make sure that plan completion listener is also properly triggered for this usage.

hiteshsharma:
@mchen07 comments addressed

Issue #205, opened 2019-01-27
Can we support @Nonnull and @Nullable annotations in public method return values and parameters?


hiteshsharma:
Can we support @Nonnull and @Nullable annotations in public method return values and parameters?
Adding these can help parseq users write accurate Parseq code more easily by getting help from IDE in null detection wherever applicable.

If agreed, I can take this task and submit a review.

mchen07:
This is a question raised before. The problem is that there are multiple versions of @nonull and @nullable annotations, as well as several static analysis tools to handle these annotations like findBugs, Android IDE, etc. When you mention IDE here, do you mean Android developer IDE? Typical IntelliJ IDE will not do null detection for you until you hook up with some static analysis checker tool.

hiteshsharma:
Yeah I agree there are multiple versions. We should stick to Java versions (`javax.annotation.Nonnull` and `Nullable`). Both IntelliJ and Eclipse have built-in support for null detection using these annotations. No plugin or stylecheck addition required.
These annotations are useful even for developers without IDE as they serve as a good documentation and set right expectations for the API.

mchen07:
How do you compare this annotation with Java Optional?

PR #206, opened 2019-02-11
add Nonnull and Nullable annotations to ContextRun function

hiteshsharma:
This change adds support for `javax.annotation.Nonnull` and `javax.annotation.Nullable` annotations for Parseq API.
This change is a follow up on discussion #205 
This change only adds annotations for `Task.contextRun` method. Once this is accepted, I'll create more changes for adding annotations to other functions.
Annotations are decided based on current code usage.
@mchen07

### DIFF
--- a/subprojects/parseq/build.gradle
+++ b/subprojects/parseq/build.gradle
@@ -35,7 +35,6 @@ configurations {
 }
 
 dependencies {
-  compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.2'
   compile group: 'org.codehaus.jackson', name: 'jackson-core-asl', version: '1.8.8'
   compile group: 'org.codehaus.jackson', name: 'jackson-mapper-asl', version: '1.8.8'
   compile group: "org.slf4j", name: "slf4j-api", version: "1.7.25"
--- a/subprojects/parseq/build.gradle
+++ b/subprojects/parseq/build.gradle
@@ -35,6 +35,7 @@ configurations {
 }
 
 dependencies {
+  compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.2'
   compile group: 'org.codehaus.jackson', name: 'jackson-core-asl', version: '1.8.8'
   compile group: 'org.codehaus.jackson', name: 'jackson-mapper-asl', version: '1.8.8'
   compile group: "org.slf4j", name: "slf4j-api", version: "1.7.25"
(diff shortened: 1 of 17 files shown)


hiteshsharma:
@mchen07 agree. I just picked the first public function in `Task.java` to demonstrate the change. I plan to enable these for entire public API one function at a time. If you think beginning with some other function would be a good idea, I can send an update.

mchen07:
I may prefer that we start with adding annotation to `Task` interface.

hiteshsharma:
@mchen07 reverted original commit and added `@Nonnull` annotations to `map` functions in `Task` interface. If this looks good, will do this for all `default` functions.
let me know what you think.

subprojects/parseq/build.gradle
```diff
@@ -35,6 +35,7 @@ configurations {
 }
 
 dependencies {
+  compile group: 'com.google.code.findbugs', name: 'jsr305', version: '3.0.2'
```
mchen07: why do we need to add this findbugs dependency? As we discussed earlier, I hope that this annotation should not depend on one particular static analysis tools.

hiteshsharma: `javax.annotation` package is part of findbugs. This is just lib import, we don't need to run findbugs to use these annotations but if you're not willing to import lib as well then this change can't be made.
In my opinion importing this lib should not have any impact and it's a very standard practice.
let me know what you think.

hiteshsharma: @mchen07 bump

mchen07: Sorry for the delay. I am still struggling whether we should introduce such @Nullable annotation where different IDE may have different support (for example, IntelliJ has its own @Nullable support, which is different from `javax.annotation`, see https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html), or instead we should use java Optional to indicate this, where we don't depend on the implementation of static code analysis tool.

hiteshsharma: Hey @mchen07 Intellij Documentation doesn't mentions it (probably old) but IntelliJ support it. You can go to Intellij `Preferences -> Editor -> Inspections -> Probable bugs -> @NotNull/@Nullable problems -> Configure annotations`
There you can see `javax.annotation.Nullable` and `Notnull` too

mchen07: Thanks for the tip @hiteshsharma. Have you seen this: https://github.com/spotbugs/spotbugs/pull/180? Recently spotbugs is preferred over findBugs, will this have side effect on that trend?

PR #211, opened 2019-04-22
In ParSeqRestClient align createTaskWithD2Timeout and createTaskWithT…

FreCap:
In ParSeqRestClient align createTaskWithD2Timeout and createTaskWithTimeout to always run the timeout task with the same approach.
We want to guarantee the exact same behavior in the two cases

### DIFF
--- a/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
+++ b/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
@@ -213,7 +213,7 @@ public class ParSeqRestClient extends BatchingStrategy<RequestGroup, RestRequest
 
   // check whether we need to apply timeout to a rest.li request task.
   private boolean needApplyTaskTimeout(RequestContext requestContext, ConfigValue<Long> timeout) {
-    // **return false** if no timeout configured or per-request timeout already specified in request context
+    // return false if no timeout configured or per-request timeout already specified in request context
     return timeout.getValue() != null && timeout.getValue() > 0 && !hasRequestContextTimeout(requestContext);
   }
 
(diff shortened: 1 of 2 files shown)


subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -289,7 +266,7 @@ private boolean needApplyTaskTimeout(RequestContext requestContext, ConfigValue<
     if (!taskNeedTimeout) {
       return requestTask;
     } else {
-      return withD2Timeout(requestTask, timeout);
+      return withTimeout(requestTask, timeout);
```
mchen07: Put a comment here 
// still enforce parseq client timeout if for some reason downstream services are not timed out properly.

mchen07: Also, would you mind make one javadoc change in method 
```
  // check whether we need to apply timeout to a rest.li request task.
 private boolean needApplyTaskTimeout(RequestContext requestContext, ConfigValue<Long> timeout) {
    // **return false** if no timeout configured or per-request timeout already specified in request context
     return timeout.getValue() != null && timeout.getValue() > 0 && !hasRequestContextTimeout(requestContext);
  }
```

mchen07: Update CHANGELOG.md by following our developer guide.

FreCap: Updated

subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -236,7 +213,7 @@ private boolean hasRequestContextTimeout(RequestContext requestContext) {
 
   // check whether we need to apply timeout to a rest.li request task.
   private boolean needApplyTaskTimeout(RequestContext requestContext, ConfigValue<Long> timeout) {
-    // if no timeout configured or per-request timeout already specified in request context
+    // **return false** if no timeout configured or per-request timeout already specified in request context
```
mchen07: **return false** -> return false

I used that formatter just to show you what to be changed.

FreCap: :) I'll fix it

PR #224, opened 2019-07-22
Change Regex pattern to match the lambda function description in Zing

bolobobo:
In HotSpot (java 8), the dynamic identifier for lambda expression like follows: 
HelloWorld$$Lambda$1 
In Zing (java 8):
HelloWorld$$Lambda$lambda$main$0$1310938867 
In parseq-lambda-names module, it uses a regular expression to filter the lambda function identifier. Since the lambda identifier in Zing can’t match the pattern("^.*\\$\\$Lambda\\$\\d+$"), it would fail to do the following work like, eg, to find the position of that lambda function in source code. 
So change the pattern to match the dynamic identifier in Zing.

### DIFF
--- a/subprojects/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/Util.java
+++ b/subprojects/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/Util.java
@@ -12,7 +12,7 @@ import org.objectweb.asm.tree.TypeInsnNode;
 
 class Util {
 
-  private static final Pattern LAMBDA_NAME_PATTERN = Pattern.compile("^.*\\$\\$Lambda\\$\\d+$");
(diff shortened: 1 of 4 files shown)


subprojects/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/Util.java
```diff
@@ -12,7 +12,7 @@
 
 class Util {
 
-  private static final Pattern LAMBDA_NAME_PATTERN = Pattern.compile("^.*\\$\\$Lambda\\$\\d+$");
+  private static final Pattern LAMBDA_NAME_PATTERN = Pattern.compile("^.*\\$\\$Lambda\\$.*$");
```
mchen07: Any unit test on this?

PR #229, opened 2019-08-29
Enabled test utility for JUnit5

mockitoguy:
- avoided couplings with TestNG
- introduced base classes to be used by JUnit5 customers

This implementation was pair-programmed with Jiaqi Guan

### DIFF
--- a/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
+++ b/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
@@ -16,10 +16,6 @@
 
 package com.linkedin.parseq;
 
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
-
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
@@ -167,7 +163,10 @@ public class ParSeqUnitTestHelper {
   public <T> T runAndWait(final String desc, Task<T> task, long time, TimeUnit timeUnit) {
     try {
       _engine.run(task);
-      assertTrue(task.await(time, timeUnit));
+      boolean result = task.await(time, timeUnit);
+      if (!result) {
+        throw new AssertionError("Expected task result to be successful");
+      }
       return task.get();
     } catch (InterruptedException e) {
       throw new RuntimeException(e);
@@ -219,15 +218,22 @@ public class ParSeqUnitTestHelper {
       long time, TimeUnit timeUnit) {
     try {
       runAndWait(desc, task, time, timeUnit);
-      fail("An exception is expected, but the task succeeded");
-      // just to make the compiler happy, we will never get here
-      return null;
+      throw new AssertionError("An exception is expected, but the task succeeded");
     } catch (PromiseException pe) {
       Throwable cause = pe.getCause();
       assertEquals(cause.getClass(), exceptionClass);
       return exceptionClass.cast(cause);
     } finally {
       logTracingResults(desc, task);
+    }
+  }
+
+  //We don't want to use TestNG assertions to make the test utilities
+  // class useful for non TestNG users (for example, JUnit).
+  //Hence, we're writing our own private assertEquals method
+  static void assertEquals(Object o1, Object o2) {
+    if ((o1 == null && o2 != null) || (o1 != null && !o1.equals(o2))) {
+      throw new AssertionError("Object " + o1 + " is expected to be equal to object: " + o2);
     }
   }
 
--- a/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
+++ b/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2019 LinkedIn, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.linkedin.parseq.junitjupiter;
+
+import com.linkedin.parseq.AbstractBaseEngineTest;
+import com.linkedin.parseq.EngineBuilder;
+import com.linkedin.parseq.ParSeqUnitTestHelper;
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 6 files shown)


mockitoguy:
Friendly ping :)

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
```diff
@@ -0,0 +1,71 @@
...
+ * This class creates new Engine and shuts it down before and after every test method, so it can't be used
+ * to run tests in parallel.
+ *
+ * The difference between this class and {@link BaseEngineParJunitJupiterTest} is that {@code BaseEngineParJunitJupiterTest} creates new
+ * {@code Engine} instance only once for all tests in the class and thus can be used to run test methods in parallel.
+ *
+ * @see ParSeqUnitTestHelper
+ * @see BaseEngineParJunitJupiterTest
+ */
+public class BaseEngineJUnitJupiterTest extends AbstractBaseEngineTest {
```
mchen07: I myself is also debating whether we should use a different class name or the same name since you have put that in a different package. The downside for this is that when users remove `testNg` dependency, they have to not only change imports but also code. The pro is that this will be very explicit for users to think what class to use.

subprojects/parseq/src/test/java/com/linkedin/parseq/ParSeqUnitTestHelperTest.java
```diff
@@ -0,0 +1,17 @@
+package com.linkedin.parseq;
+
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertThrows;
+
+public class ParSeqUnitTestHelperTest {
```
mchen07: Can we have a test for `ParSeqUnitTestHelper` using Junit5?

mchen07: Another maybe a quite dumb question: I know that the change you made in `ParSeqUnitTestHelper` is to remove compile dependency on `testNg`. Even without this change, currently consumers using JUNit can still use `ParSeqUnitTestHelper` for their parseq task. So your change is mainly to avoid pulling testNg dependency transitively for consumers using JUnit, right?

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
```diff
@@ -231,6 +228,15 @@ public ScheduledExecutorService getScheduler() {
     }
   }
 
+  //We don't want to use TestNG assertions to make the test utilities
+  // class useful for non TestNG users (for example, JUnit).
+  //Hence, we're writing our own private assertEquals method
+  static void assertEquals(Object o1, Object o2) {
```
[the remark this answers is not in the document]
mockitoguy: It's a package protected method, not exposed to clients. This class is not actively developed so it won't have new, sophisticated assertions in the future. Even if it has, we can always take in AssertJ as dependency. However today, given minimum amount of assertions, and that the class is not growing, it's best to avoid a dependency

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
mockitoguy: I suggest an explicit, different name. The migration argument is not strong enough: search replace across files is easy.

Happy to change the name to be the same if that's the team's decision. I don't mind either way.

subprojects/parseq/src/test/java/com/linkedin/parseq/ParSeqUnitTestHelperTest.java
mockitoguy: Problem is that it won't be picked up by the build. For the entire "parseq" module we should use one testing framework.

It's fine to continue using TestNG for parseq project. However, we should make sure that parseq customers can use JUnit (hence this PR :-D)

bolobobo: What is this class used for?

mockitoguy: Yes. We need to stay compatible for existing customers, and offer clean classpath to new customer who use JUnit5+

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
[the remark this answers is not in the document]
mchen07: Got it. It is package protected, so we are good on this.

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
mchen07: Ok, you convinced me.

subprojects/parseq/src/test/java/com/linkedin/parseq/ParSeqUnitTestHelperTest.java
mockitoguy: It's a unit test for "ParSeqUnitTestHelper" class

mockitoguy:
Friendly ping :)

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private volatile boolean _setUpCalled = false;
+  private volatile boolean _tearDownCalled = false;
+
+  @BeforeEach
+  public void setUpBaseEngineTest() throws Exception {
+    if (!_setUpCalled) {
```
GuiForget: Why not use `@BeforeAll` annotation combined with `@TestInstance(Lifecycle.PER_CLASS)`?

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineJUnitJupiterTest.java
```diff
@@ -0,0 +1,71 @@
...
+      _setUpCalled = true;
+      _tearDownCalled = false;
+      getParSeqUnitTestHelper().setUp();
+    }
+  }
+
+  @AfterEach
+  public void tearDownBaseEngineTest() throws Exception {
```
GuiForget: I may be wrong but I think using @AfterEach when running tests in parallel will cause some issues here.

bolobobo: @GuiForget In your case, you may need to use https://github.com/linkedin/parseq/blob/master/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/junitjupiter/BaseEngineParJunitJupiterTest.java

GuiForget: Ah I hadn't seen that other class. Thank you

PR #236, opened 2019-10-20
Fixed 19 build time warnings

hiteshsharma:
See all the warnings here
https://scans.gradle.com/s/n3ap5h46hu6ne/console-log#L84

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Engine.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Engine.java
@@ -289,8 +289,8 @@ public class Engine {
    * This method throws {@code IllegalStateException} if Engine does not have capacity to run the task.
    * Engine's capacity is specified by a {@value #MAX_CONCURRENT_PLANS} configuration property. Use
    * {@link EngineBuilder#setEngineProperty(String, Object)} to set this property.
-   * For the sake of backwards compatibility default value for a {@value #MAX_CONCURRENT_PLANS} is
-   * {@value #DEFAULT_MAX_CONCURRENT_PLANS} which essentially means "unbounded capacity".
+   * For the sake of backwards compatibility default value for a {@link #MAX_CONCURRENT_PLANS} is
+   * {@link #DEFAULT_MAX_CONCURRENT_PLANS} which essentially means "unbounded capacity".
    *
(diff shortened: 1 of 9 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Engine.java
```diff
@@ -58,7 +58,7 @@
   private static final int DEFUALT_MAX_RELATIONSHIPS_PER_TRACE = 65536;
 
   public static final String MAX_CONCURRENT_PLANS = "_MaxConcurrentPlans_";
-  private static final int DEFUALT_MAX_CONCURRENT_PLANS = Integer.MAX_VALUE;
+  private static final int DEFAULT_MAX_CONCURRENT_PLANS = Integer.MAX_VALUE;
```
mchen07: Is this already in another PR?

subprojects/parseq/src/main/java/com/linkedin/parseq/internal/PlanDeactivationListener.java
```diff
@@ -5,8 +5,6 @@
 /**
  * This listener interface allows receiving notifications about plan deactivation.
  * It can be used for monitoring and to optimize task implementation e.g. to implement batching.
- * See {@link EngineBuilder#setPlanActivityListener(PlanDeactivationListener)}
```
mchen07: Why do you remove this? This method is there in `EngineBuilder`.

subprojects/parseq/src/main/java/com/linkedin/parseq/Engine.java
hiteshsharma: Deleted that PR. This should be the only one with this change now

subprojects/parseq/src/main/java/com/linkedin/parseq/internal/PlanDeactivationListener.java
hiteshsharma: This method was replaced by `setPlanDeactivationListener` in https://github.com/linkedin/parseq/commit/bfcc5ffb2680fc36e9a7bf4f139f8df479bb2b4d but this comment never got changed. I wasn't able to find this method that's why I removed it but let me replace it with the new one

PR #238, opened 2019-11-14
Making sure _taskDoneLatch doesn't keep growing

darthsidpal:
The _taskDoneLatch map is ever growing as we keep running more and more tasks. If the CountDownLatch is 0, I think it means no one else is using it and so remove it from the map.

### DIFF
--- a/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
+++ b/subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
@@ -325,6 +325,10 @@ public class ParSeqUnitTestHelper {
     public void onPlanCompleted(PlanContext planContext) {
       CountDownLatch latch = _taskDoneLatch.computeIfAbsent(planContext.getRootTask(), key -> new CountDownLatch(1));
       latch.countDown();
+      
+      if (latch.getCount() == 0L) {
+        _taskDoneLatch.remove(planContext.getRootTask());
+      }
     }
 
     public void await(Task<?> root, long timeout, TimeUnit unit) throws InterruptedException {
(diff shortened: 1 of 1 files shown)


subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
```diff
@@ -325,6 +325,10 @@ public int countTasks(Trace trace) {
     public void onPlanCompleted(PlanContext planContext) {
       CountDownLatch latch = _taskDoneLatch.computeIfAbsent(planContext.getRootTask(), key -> new CountDownLatch(1));
       latch.countDown();
+      
+      if (latch.getCount() == 0L) {
```
mchen07: This fix may not work. This call is called, but some tests may need to call `await` .

darthsidpal: from what I understand, this onPlanCompleted is called by the engine when it is done.
Are you suggesting that even in await, if latch.await returns false- that is the work wasn't done, we forcibly remove the latch?

mchen07: Capture what we have discussed offline here. We need to probably handle two cases to make sure that `runAndWaitForPlanComplete` API in ParSeqUnitTestHelper not break.
1. TaskDoneListener.await is called before TaskDoneListener.onPlanCompleted.
2. TaskDoneListener.await is called after TaskDoneListener.onPlanCompleted.

subprojects/parseq/src/test-utils/java/com/linkedin/parseq/ParSeqUnitTestHelper.java
```diff
@@ -332,8 +333,17 @@ public void onPlanCompleted(PlanContext planContext) {
     }
 
     public void await(Task<?> root, long timeout, TimeUnit unit) throws InterruptedException {
-      CountDownLatch latch = _taskDoneLatch.computeIfAbsent(root, key -> new CountDownLatch(1));
-      latch.await(timeout, unit);
+      CountDownLatch latch = _taskDoneLatch.get(root);
+      
+      // If the latch is null, it means that onPlanCompleted was already called which removed the latch.
```
mchen07, line 338: Put some comment saying that this API is supposed to be called together with `setupAwait`. Or you can put some javadoc on`setupAwait` to explain this. Otherwise, it is easy to forget the sequence we are assuming here.

darthsidpal: added! :)

Issue #239, opened 2020-01-16
ParSeq doesn't complete task if error during logging


mchen07:
When a ParSeq plan fails with an exception, ParSeq attempts to log the exception. If converting the exception into a string fails with another exception, the final Task is not actually completed, causing the ParSeq plan to stall indefinitely.

PR #240, opened 2020-01-16
Guard exception thrown from tracing to avoid unfinished plan.

mchen07:
Fix for https://github.com/linkedin/parseq/issues/239

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/BaseTask.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/BaseTask.java
@@ -252,8 +252,10 @@ public abstract class BaseTask<T> extends DelegatingPromise<T>implements Task<T>
         traceFailure(reason);
       } catch (Throwable ex) {
         LOGGER.warn("Exception thrown in logging trace for failure!", ex);
+      } finally {
+        // guard any exception that may throw from catch block
+        getSettableDelegate().fail(reason);
       }
-      getSettableDelegate().fail(reason);
       return true;
     }
     return false;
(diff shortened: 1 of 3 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/BaseTask.java
```diff
@@ -248,7 +248,11 @@ public String getName() {
   public boolean cancel(final Exception rootReason) {
     if (transitionCancel(rootReason)) {
       final Exception reason = new CancellationException(rootReason);
-      traceFailure(reason);
+      try {
+        traceFailure(reason);
+      } catch (Throwable ex) {
+        LOGGER.warn("Exception thrown in logging trace for failure!", ex);
```
avik-das: The issue we found was when the original `reason` could not be logged due to an error in its `toString`. Do we have any concern that logging the _new_ error might have the same issue? Would it be better to have two stages to this?

1. Catch the logging error, then try to log that error.
2. Catch _that_ error if it happens, and log a generic, static message.

subprojects/parseq/src/test/java/com/linkedin/parseq/TestTasks.java
```diff
@@ -48,6 +48,13 @@
  */
 public class TestTasks extends BaseEngineTest {
 
+  private static class UnloggableException extends Exception {
+    @Override
+    public String toString() {
+      throw new RuntimeException("Cannot log an UnloggableException!");
```
karthikbalasub: Change this to UnloggableException to test the scenario Avik mentioned?

avik-das: Yeah, that would do the trick.

Optional: successfully logging the exception thrown by `toString` vs. falling back to a second `catch` block are technically two different scenarios. We don't distinguish between them in the tests because we don't check what actually gets logged, but they do execute different code paths. I would recommend having two test cases:

- The existing `UnloggableException`.
- A `ReallyUnloggableException` where the `toString` throws a `ReallyUnloggableException`.

mchen07: Sounds good. Thanks for the suggestion.

mchen07: Through testing, Logger.warn("...", ex) for an exception that can throw exception in `toString` will not throw exception for some logger. To be safe, I used finally block and added another test for that.

avik-das: If the logging of the second error doesn't actually throw a third error, then I don't think `ReallyUnloggableException` actually exercises the `catch` block. But, without making the logger an injectable dependency, I don't think we even _can_ exercise the `catch`. Still, I think the additional test case is useful, so we should still leave it in.

PR #248, opened 2020-04-15
Move test artifacts to the new ':parseq-test-api' sub-project

evanw555:
Discontinues publishing test artifacts for the `:parseq` sub-project, and instead publishes them as main artifacts for the new `:parseq-test-api` sub-project. The motivation for this is to follow Maven best practices by avoiding test artifacts being included in the consumer-facing publication.

To be more specific, the artifact `parseq-3.0.13-test.jar` (which can currently be referenced via the `testArtifact` configuration) will instead be `parseq-test-api-4.0.0.jar` (and be referenced via standard configurations e.g. `testCompile`). The extra test artifacts (`*-test-sources.jar`, `*-test-javadoc.jar`) will also be published as normal Maven artifacts (`*-sources.jar`, `*-javadoc.jar`).

**Note:**
* This is a backward-incompatible change and necessitates a major version bump.
* Ignore all changes to `*.java` files. All I did was move them to the new sub-project, resulting in automatic import optimization for some of them.


DPUkyle:
Slight note on documentation:
> and be referenced via standard configurations e.g. `compile`

Since the new subproject is a testing API, it sounds more natural to me if you mention `testCompile` or `testImplementation` configurations, not `compile`.

evanw555:
> nit: we may want to add a comment on classes moved in 3.* and release a patch version before releasing this major version. upto you
@aman1309 Could you please clarify? If I understand you correctly, you're saying we should release one last patch version in 3.* which somehow informs the user of the change. Since Parseq doesn't provide a plugin like Pegasus does, how would we do this?

aman1309:
> > nit: we may want to add a comment on classes moved in 3.* and release a patch version before releasing this major version. upto you
> > @aman1309 Could you please clarify? If I understand you correctly, you're saying we should release one last patch version in 3.* which somehow informs the user of the change. Since Parseq doesn't provide a plugin like Pegasus does, how would we do this?

I was thinking probably add a comment in classes Javadoc but could be a overkill. Feel free to ignore it

settings.gradle
```diff
@@ -1,6 +1,7 @@
 rootProject.name = 'parseq'
 
 def modules = [ /* the name of the modules to use */
+    'parseq', // core part code, including test utility
```
mchen07, line 4: Does this still include test utility? The comment seems indicating that.

evanw555: True, this is no longer accurate. I'll fix this

mockitoguy:
Looks great! Much simpler! Thank you!

PR #253, opened 2020-04-21
Add testArtifact configuration back

evanw555:
Temporary fix to maintain compatibility with transitive consumers who
are still explicitly depending on the testArtifact configuration of
the parseq module. This simply tacks on the parseq-test-api dependency.

Steps to reproduce error:
1. Consumer A depends directly on `parseq-test-api@4.0.0`
2. Consumer B depends directly on Consumer A, yet also depends directly on `parseq@3.*` using the `testArtifact` configuration.
3. Consumer B's dependency on `parseq` will be bumped up to `4.0.0`, which will cause an error because the `testArtifact` configuration doesn't exist for `parseq@4.0.0`.

Without this fix, transitive consumers of parseq 4.0.0 may still be explicitly depending on the `testArtifact` configuration of the `parseq` module, which would result in an error like this:

```
Could not resolve all files for configuration ':consumer:testCompileClasspath'.
   > Could not resolve com.linkedin.parseq:parseq:4.0.0-SNAPSHOT.
Required by:
    project :consumer
      > Project :consumer declares a dependency from configuration 'testCompile' to configuration 'testArtifact' which is not declared in the descriptor for com.linkedin.parseq:parseq:4.0.0-SNAPSHOT.
```

### DIFF
--- a/subprojects/parseq/build.gradle
+++ b/subprojects/parseq/build.gradle
@@ -23,6 +23,10 @@ configurations {
     description "bootstrapping dependencies"
     visible = false
   }
+  // Maintains transitive consumer compatibility for consumers of parseq 3. TODO: remove in parseq 5
+  testArtifact {
+    description "this is deprecated and will be removed, please depend on parseq-test-api instead"
+  }
 }
 
 dependencies {
@@ -34,6 +38,9 @@ dependencies {
   testCompile project(':parseq-test-api')
   testCompile group: "org.testng", name: "testng", version: "6.9.9"
   testCompile group: "org.slf4j", name: "slf4j-log4j12", version: "1.7.25"
+
+  // TODO: remove in parseq 5
+  testArtifact project(':parseq-test-api')
 }
 
 // publish to maven
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 v4.0.1
 ------
 
+* Added the `testArtifact` configuration back for the `:parseq` module as a temporary fix to maintain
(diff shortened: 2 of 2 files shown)


subprojects/parseq/build.gradle
```diff
@@ -34,6 +36,9 @@ dependencies {
   testCompile project(':parseq-test-api')
   testCompile group: "org.testng", name: "testng", version: "6.9.9"
   testCompile group: "org.slf4j", name: "slf4j-log4j12", version: "1.7.25"
+
+  // TODO: remove in parseq 5
```
mchen07, line 40: How can you remove in parseq 5? Will the same problem persist?

evanw555: Answered this offline, but I'll repeat here for visibility. Yes, removing the configuration in parseq 5 will cause the same problem for anyone who's still consuming the `testArtifact` configuration. However, this gives us a window of support between parseq 4 and 5 where support for both usage patterns is supported, which makes it easier for us to migrate all internal consumers. By the time we want to bump to parseq 5, all internal consumers will have been migrated away from the `testArtifact` configuration, so the impact will be much more manageable.

evanw555: Perhaps it would've been wiser to bump the minor version first (since the configuration wouldn't have been removed yet), then bump the major version to remove the configuration. Hindsight is always 20/20 though... Anyway, the goal here is to provide a window of support for both options. We want to remove the `testArtifact` configuration as soon as we can though.

mockitoguy:
Looks great. Thank you!

PR #258, opened 2020-04-24
fix mac build issue for tracevis

aman1309:
Upgrading phantomjs to solve one of the build issue in testCov, fails on mac only since c++ libs used to build are different on mac. Other issues relating to installation failure:

Root cause is contextify being no longer support as it is merged native to node. 
Permanent fix is upgrade node and package dependencies. 

for now will have use below patches(one time). I will add specific cmds to developer guide.

Refer:
https://github.com/nodejs/node-gyp/issues/569
https://github.com/nodejs/node-gyp/issues/469

### DIFF
--- a/subprojects/parseq-tracevis/package.json
+++ b/subprojects/parseq-tracevis/package.json
@@ -9,7 +9,7 @@
     "jshint": "~2.1.11",
     "istanbul": "~0.1.44",
     "mocha": "1.2.x",
-    "phantomjs": "~1.9.2-2",
+    "phantomjs": "^2.1.7",
     "uglify-js": "2.6.0"
   },
   "dependencies": {


subprojects/parseq-tracevis/package.json
```diff
@@ -9,7 +9,7 @@
     "jshint": "~2.1.11",
     "istanbul": "~0.1.44",
     "mocha": "1.2.x",
-    "phantomjs": "~1.9.2-2",
+    "phantomjs": "^2.1.7",
```
mchen07, line 12: For my own understanding, what does ^  mean here? 2.1.7 or above?

aman1309: equivalent to 2.*
~2.1.7  is 2.1.*

mchen07: I see. Thanks.

evanw555:
I would advise that you leave a note in the `README.md` file for `:parseq-tracevis` showing the exact commands needed to fix this. Not everyone has access to our internal development guide, plus this would make the information easier to access for anyone trying to build parseq.

aman1309:
> I would advise that you leave a note in the `README.md` file for `:parseq-tracevis` showing the exact commands needed to fix this. Not everyone has access to our internal development guide, plus this would make the information easier to access for anyone trying to build parseq.

I think we should add a dev guide which can include this, will add this task in queue

PR #259, opened 2020-04-28
Set up automated releases, improve helper scripts

evanw555:
Sets up automatic Bintray publications in Travis whenever a tag is pushed. Improves the `local-release` script and re-purposes the `release` script to create and push a tag to trigger a deployment. Also removes the "Built-By" property from the JAR manifest.

**Release Workflow:**
1. `./scripts/release` to release from the current HEAD.
2. Go to ParSeq on GitHub to verify that the publication job ran successfully.

**Snapshot Workflow:**
- `./scripts/local-release` to publish `x.y.z` to `~/local-repo`
- `./scripts/local-release -s` to publish `x.y.z-SNAPSHOT` to `~/local-repo`

**Note:**
- I did some mocked testing of the `publish-tag.sh` Travis script, but I can't do an end-to-end test until this is merged.
- Both scripts (`release` and `local-release`) support the `-h, --help` argument, so they should be pretty user friendly.

**Follow-up Items:**
- Document how to perform a hotfix release.
- See how email notifications work, address any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,30 @@
 language: java
-
-script: ./gradlew build
-
 jdk:
   - oraclejdk8
 
+# Anything after trusty doesn't have Java 8
 dist: trusty
 
-#Below skips the installation step completely (https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step)
-#If we don't skip it Travis runs unnecessary Gradle tasks like './gradlew assemble'
-install:
- - true
+# Below skips the installation step completely (https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step)
+# If we don't skip it Travis runs unnecessary Gradle tasks like './gradlew assemble'
+install: true
 
-# Only build on master and release tags
+# Build and test each commit on master and each PR to master
+script: ./gradlew build
 branches:
   only:
   - master
-  - /^v\d+\.\d+\.\d+$/
+
+# Publish a new version on tag push
+deploy:
+  provider: script
+  script: scripts/travis/publish-tag.sh
+  on:
+    tags: true
+
+# Send email notifications to the user associated with a tag deployment
+notifications:
+  email:
+    if: tag IS present
+    on_failure: always
+    on_success: always
--- a/scripts/local-release
+++ b/scripts/local-release
@@ -1,27 +1,50 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # The purpose of this script is to publish artifacts to ~/local-repo
 
-if [ ! -f gradle.properties ]; then
-  echo 'Could not find gradle.properties. Run this command from the root of the project.'
-  exit 1
+# Ensure that the script is being run from the root project directory
+PROPERTIES_FILE='gradle.properties'
+if [ ! -f "$PROPERTIES_FILE" ]; then
+  echo "Could not find $PROPERTIES_FILE, please run this script from the root project directory."
+  exit 2
 fi
+
+# Process CLI arguments
+EXTRA_PROPERTY='-Prelease'
+for ARG in "$@"; do
+  if [ "$ARG" = '-h' ] || [ "$ARG" = '--help' ]; then
+    cat ./scripts/help-text/local-release.txt
+    exit 0
+  elif [ "$ARG" = '-s' ] || [ "$ARG" = '--snapshot' ]; then
+    EXTRA_PROPERTY='-Psnapshot'
+  else
+    echo "Unrecognized option: $ARG"
+    echo ''
+    cat ./scripts/help-text/local-release.txt
+    exit 2
+  fi
+done
 
 if [ ! -d "$HOME" ]; then
   echo 'Cannot perform local release, $HOME is not set to a valid directory.'
   exit 1
 fi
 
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,19 +1,30 @@
...
-  - /^v\d+\.\d+\.\d+$/
+
+# Publish a new version on tag push
+deploy:
+  provider: script
+  script: scripts/travis/publish-tag.sh
+  on:
+    tags: true
+
+# Send email notifications to the user associated with a tag deployment
```
aman1309, line 25: will this send a notification to all contributors? or how are users associated with a tag determined?

scripts/local-release
```diff
@@ -1,27 +1,50 @@
...
+VERSION=$(awk 'BEGIN { FS = "=" }; $1 == "version" { print $2 }' $PROPERTIES_FILE | awk '{ print $1 }')
+if [ "$EXTRA_PROPERTY" = '-Psnapshot' ] && [[ "$VERSION" != *-SNAPSHOT ]]; then
+  VERSION="${VERSION}-SNAPSHOT"
+fi
+
 echo "Publishing parseq $VERSION to ${LOCAL_REPO}..."
 
 # Publish artifacts to Maven local, but override the repo path as ~/local-repo
-./gradlew -Dmaven.repo.local=$LOCAL_REPO -Prelease publishReleasePublicationToMavenLocal
+./gradlew -Dmaven.repo.local=$LOCAL_REPO $EXTRA_PROPERTY publishReleasePublicationToMavenLocal
```
aman1309, line 47: should we hard code override here? request you to consider override directory as an argument or using system env variable to override.

.travis.yml
evanw555: Nope. According to the [Travis docs](https://docs.travis-ci.com/user/notifications/#how-is-the-build-email-receiver-determined), the recipient is determined using the email associated with the "committer and author", which presumably refers to the commit committer and author. I'm not sure how this will play out for tags; will it choose the commit's author, or the tagger of the tag? Can't say for now. We can also change it to send to an internal mailing list.

scripts/local-release
evanw555: For our standard use case, we will almost always just need `~/local-repo`. If someone really needs to publish somewhere else locally, they can just directly call the command. I can look into it to see if it's easy to add this as another script argument.

scripts/travis/publish-tag.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+# Ensure the commit environment variable exists
+if [ -z "$TRAVIS_COMMIT" ]; then
+  echo 'Cannot find environment variable named TRAVIS_COMMIT, did the Travis API change?'
+  exit 1
+fi
+
+# Ensure that the tag commit is an ancestor of master
+git merge-base --is-ancestor $TRAVIS_COMMIT master
```
aman1309, line 45: what about hotfix release? should those be done manually from local?
It might be good to add this script on any release tag or have a task for adding a hotfix script, so that we don't need to revert all changes after bad code.

evanw555: I think that case will have to be done manually on the command line. It'll be a rare case and I think having more stringent checks in the git tag deployment is more important than improving the dev experience of this rare corner case. Thanks for pointing this out though, this is something that will need to be documented in the updated dev guide.

scripts/local-release
evanw555: Based on our conversation offline, you make some good arguments as to why this could be useful to add as an extra argument, but it would require some more work right now. Can we agree to defer this until later? I can add a TODO

aman1309: sure

subprojects/parseq-benchmark/build.gradle
```diff
@@ -16,7 +16,6 @@ task fatJar(type: Jar) {
   manifest {
     attributes("Created-By": "Gradle",
         "Version": version,
-        "Built-By": project.rootProject.ext.builtByUser,
```
nickibi, line 19: Curious about why do we remove this one?

subprojects/parseq-tracevis-server/build.gradle
```diff
@@ -37,7 +37,6 @@ task fatJar(type: Jar, dependsOn: ':parseq-tracevis:makeDist') {
   manifest {
     attributes("Created-By": "Gradle",
         "Version": version,
-        "Built-By": project.rootProject.ext.builtByUser,
```
nickibi, line 40: Same here.

subprojects/parseq-benchmark/build.gradle
evanw555: According to Szczepan, this sort of "volatile" metadata can harm cacheability. See: https://github.com/linkedin/parseq/pull/243#discussion_r404343718

Also, since all the publications will be done from Travis using a certain person's Bintray credentials, it doesn't seem appropriate.

subprojects/parseq-tracevis-server/build.gradle
evanw555: See comment above.

subprojects/parseq-benchmark/build.gradle
nickibi: Good to know that.

PR #262, opened 2020-04-30
Include tags in Travis build, set env variables

evanw555:
Previously removed the tags from the build condition, but turns out that actually removes it from the deploy condition as well.

Will attempt to release `4.1.1` from this commit.


evanw555:
Failure is a familiar flaky test. Going to merge anyway.

Issue #244, opened 2020-04-06
Use long instead Long in ShallowTraceBuilder to avoid memory over-use for parseq tracing.


mchen07:
Currently `ShallowTraceBuilder` is using `Long` for `_id`, `_startNanos`, `_endNanos`, `_pendingNanos`. When a plan involves too many tasks (both system and user-defined), we will have too many instances of `ShallowTraceBuilder`, which leads to memory over-use. We should optimize the code to see if they can be replaced by `long`.

mchen07:
When we make this optimization, please keep public API unchanged to avoid breaking any open-source users.

junchuanwang:
This issue is closed since change pushed by @aman1309  [PR#267](https://github.com/linkedin/parseq/pull/267)

PR #267, opened 2020-05-22
use native long instead of Long wrapper for ShallowTrace

aman1309:
Solves https://github.com/linkedin/parseq/issues/244

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/trace/ShallowTraceBuilder.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/trace/ShallowTraceBuilder.java
@@ -125,7 +125,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #setNativeStartNanos(long)} instead
+   * @deprecated Use {@link #setNativeStartNanos(long)} instead
    */
   @Deprecated
   public ShallowTraceBuilder setStartNanos(Long startNanos) {
@@ -137,7 +137,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #setNativePendingNanos(long)} instead
+   * @deprecated Use {@link #setNativePendingNanos(long)} instead
    */
   @Deprecated
   public ShallowTraceBuilder setPendingNanos(Long pendingNanos) {
@@ -149,7 +149,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #setNativeEndNanos(long)} instead
+   * @deprecated Use {@link #setNativeEndNanos(long)} instead
    */
   @Deprecated
   public ShallowTraceBuilder setEndNanos(Long endNanos) {
@@ -181,7 +181,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #getNativeId()} instead
+   * @deprecated Use {@link #getNativeId()} instead
    */
   @Deprecated
   public Long getId() {
@@ -208,7 +208,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #getNativeStartNanos()} instead
+   * @deprecated Use {@link #getNativeStartNanos()} instead
    */
   @Deprecated
   public Long getStartNanos() {
@@ -219,7 +219,7 @@ public class ShallowTraceBuilder {
   }
 
   /**
-   * Deprecated. Use {@link #getNativePendingNanos()} instead
+   * @deprecated Use {@link #getNativePendingNanos()} instead
    */
   @Deprecated
   public Long getPendingNanos() {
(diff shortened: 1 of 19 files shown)


aman1309:
> Did we see any performance improvement in running benchmark profiles with this change?

There is a slight increase tried 3 times on Mac with similar conditions
Before:
Throughput: 999.362 plans/s, 300807.976 tasks/s

After:
Throughput: 1001.011 plans/s, 301304.197 tasks/s

mchen07:
> > Did we see any performance improvement in running benchmark profiles with this change?
> 
> There is a slight increase tried 3 times on Mac with similar conditions
> Before:
> Throughput: 999.362 plans/s, 300807.976 tasks/s
> 
> After:
> Throughput: 1001.011 plans/s, 301304.197 tasks/s

The initial motivation for this change is memory over-use from our ShallowTrace data structure, is it possible to measure the memory usage before and after?

subprojects/parseq/src/test/java/com/linkedin/parseq/TestTaskStates.java
```diff
@@ -268,8 +268,8 @@ private void runTask(final Task<?> task) {
   private void assertInitOrScheduled(final Task<?> task) {
     assertFalse(task.isDone());
     assertFalse(task.isFailed());
-    assertNull(task.getShallowTrace().getStartNanos());
-    assertNull(task.getShallowTrace().getEndNanos());
+    assertEquals(task.getShallowTrace().getNativeStartNanos(), -1);
```
mchen07, line 271: We should create new tests for new APIs to make sure that our change will not break existing users.

junchuanwang:
+1 for memory footprint. Can we do a heap dump?

subprojects/parseq/src/test/java/com/linkedin/parseq/TestTaskStates.java
aman1309: to keep it clean, I had tested these Uts with deprecated methods. added new tests for deprecated methods separately. Check TestShallowTraceBuilder#testDeprecatedTimeMethods

aman1309:
> > > Did we see any performance improvement in running benchmark profiles with this change?
> > 
> > 
> > There is a slight increase tried 3 times on Mac with similar conditions
> > Before:
> > Throughput: 999.362 plans/s, 300807.976 tasks/s
> > After:
> > Throughput: 1001.011 plans/s, 301304.197 tasks/s
> 
> The initial motivation for this change is memory over-use from our ShallowTrace data structure, is it possible to measure the memory usage before and after?

As you can minor GC is clean up more items before due to wrapper reference used Long
image

subprojects/parseq/src/main/java/com/linkedin/parseq/trace/ShallowTraceBuilder.java
```diff
@@ -101,31 +110,58 @@ public ShallowTraceBuilder setResultType(final ResultType resultType) {
...
-      if (_endNanos == null) {
+      if (_endNanos == -1) {
         _endNanos = nanoTime;
       }
     }
     _resultType = resultType;
     return this;
   }
 
+  @Deprecated
```
mchen07, line 127: As good practice. put javadoc 

`@deprecated replaced by ...`

subprojects/parseq/src/test/java/com/linkedin/parseq/trace/TestShallowTraceBuilder.java
```diff
@@ -97,27 +104,56 @@ public void testEquals() {
...
-    builderCopy.setEndNanos(345L);
+    builderCopy.setNativeStartNanos(123L);
+    builderCopy.setNativePendingNanos(234L);
+    builderCopy.setNativeEndNanos(345L);
     builderCopy.setResultType(ResultType.ERROR);
     assertFalse(builder.build().equals(builderCopy.build()));
   }
+
+  @Test
+  public void testDeprecatedTimeMethods() {
```
mchen07, line 132: Let's cover several permutations for those fields we touched.
1. set and get all using deprecated methods
2. set using new method, and get using deprecated method
3. set using deprecated method, and get using new method
4. set and get all using new methods (which should already covered by all other tests you have fixed)

subprojects/parseq/src/main/java/com/linkedin/parseq/trace/ShallowTraceBuilder.java
aman1309: yup. missed here. added now

subprojects/parseq/src/test/java/com/linkedin/parseq/trace/TestShallowTraceBuilder.java
aman1309: updated

subprojects/parseq/src/main/java/com/linkedin/parseq/trace/ShallowTraceBuilder.java
```diff
@@ -124,6 +124,9 @@ public ShallowTraceBuilder setResultType(final ResultType resultType) {
     return this;
   }
 
+  /**
+   * Deprecated. Use {@link #setNativeStartNanos(long)} instead
```
mchen07, line 128: we need to use @deprecated to generate proper javadoc for this.https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/deprecation/deprecation.html

PR #273, opened 2020-06-17
Wrapping Client request in try catch to propagate failure to Task

logancarmody:
Currently, if the Client implementation throws an uncaught exception, the callback isn't executed. This is to ensure the Task is always failed if the Client implementation throws an exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
-v4.1.4
+v4.1.5
 ------
+
+v4.1.4
+------  
+
+* Catching exceptions thrown by Rest.li client impl and failing the Task.
 
 v4.1.3
 ------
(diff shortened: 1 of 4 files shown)


subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -119,7 +119,11 @@ public ParSeqRestClient(final RestClient client) {
   @Deprecated
   public <T> Promise<Response<T>> sendRequest(final Request<T> request, final RequestContext requestContext) {
     final SettablePromise<Response<T>> promise = Promises.settable();
-    _client.sendRequest(request, requestContext, new PromiseCallbackAdapter<T>(promise));
+    try {
+      _client.sendRequest(request, requestContext, new PromiseCallbackAdapter<T>(promise));
+    } catch (Exception e) {
```
karthikrg, line 124: Do you want to catch Exception or throwable?

logancarmody: I guess throwable makes more sense. I can update

mchen07: +1 on Throwable

mchen07: +1 on Throwable.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 v4.1.4
```
aman1309, line 1: please add next version here

PR #277, opened 2020-07-25
Introduce parseq-all meta-project

evanw555:
Adds a new module, `parseq-all`, which is simply a meta-project that
programmatically depends on all other consumer-facing modules. This
makes it easier for certain use cases to pull in the entire dependency
tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
+v4.1.6
+------
+
 v4.1.5
 ------
+
+* Introduce `parseq-all` meta-project, which is useful for computing ParSeq's entire dependency tree.
 
 v4.1.4
 ------  
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
-version=4.1.4
+version=4.1.5
 group=com.linkedin.parseq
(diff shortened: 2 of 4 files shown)


junchuanwang:
> Adds a new module, `parseq-all`, which is simply a meta-project that
> programmatically depends on all other consumer-facing modules. This
> makes it easier for certain use cases to pull in the entire dependency
> tree.

Please describe the detailed use case where this meta project is needed in this PR, and also add it to comments in `build.gradle`

evanw555:
> so that the entire dependency tree of parseq may be pulled in easily
@junchuanwang This is the use case described in `build.gradle` without being too specific about LI-specific technology.

junchuanwang:
> > so that the entire dependency tree of parseq may be pulled in easily
> 
> @junchuanwang This is the use case described in `build.gradle` without being too specific about LI-specific technology.

I understand this is your purpose (since you did something similar to Rest.Li project), but when I was looking at the PR initially, I just found it lack of context if looking purely at the doc comments. For example if this is the first time I was reading the code,I will have questions such as why this subproject is essential and why there is no other alternatives

PR #280, opened 2020-08-06
Adding cookies to the generated request from GetRequestGroup

logancarmody:
Currently, the GetRequestGroup implementation leaves out cookies while batching requests. This causes issues, especially for batched requests to frontends which may use cookies for authentication logic.

Likely the assumption was the cookies were included in the headers, but unfortunately they are not in the R2 Request object.

This change might be backwards incompatible for some requests. It will not change the ability for requests to be batched, because currently the key in the map uses the whole request object to check for equality, so it is already checking for cookie equality. However we aren't currently sending the cookies with the batch-get request, so if that could transform the response, then this would be backwards incompatible.

### DIFF
--- a/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
+++ b/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
@@ -16,10 +16,13 @@
 
 package com.linkedin.restli.client;
 
+import java.net.HttpCookie;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/GetRequestGroup.java
```diff
@@ -434,6 +445,11 @@ public boolean equals(Object obj) {
         return false;
     } else if (!_headers.equals(other._headers))
       return false;
+    if (_cookies == null) {
```
karthikrg, line 448: You can use Objects.equals here (https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-)

logancarmody: Good call. Refactored

BrianPin:
Hi Logan,

Would it be safe to add a configuration for this feature?

By default the feature of this change can be disabled.

Only the cases you mentioned can turn on it.

logancarmody:
Sure would be happy to add a flag.

logancarmody:
On second thought, I think this is a bug and introducing a boolean flag here is very semantically weird.

We would have to surface it on the ParseqRestClient level, which would be a flag which is `_shouldAddCookiesToBatchedRequests`.

Additionally, cookies can impact behavior server-side and are already being used to determine whether to batch the requests, so it would be incorrect to expose a configuration to not attach them

PR #282, opened 2020-08-19
Reword javadoc in Task for .recover() and .withTimeout()

junchuanwang:
Reworded the javadoc for Task methods so failure handling for timeout task will be clearer.

In short

- Timeout task will be considered failure and will be handled by failure handling methods
- Cancelled task will not be considered as failure for failure handling methods
- Initiating task chain with timeout methods will return a "withtimeout" subtask. When the "withtimeout" subtask timeout, the subtask is not cancelled, but the initiating task is cancelled (this brought some confusion in the doc)

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -609,7 +609,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
 
   /**
    * Creates a new task that will handle failure of this task.
-   * Early completion due to cancellation is not considered to be a failure.
+   * Early completion due to cancellation is not considered to be a failure so it will not be recovered..
    * If this task completes successfully, then recovery function is not called.
    * <blockquote><pre>
    *
@@ -626,7 +626,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * </pre></blockquote>
    * <img src="doc-files/recover-1.png" height="90" width="462"/>
    * <p>
-   * Note that task cancellation is not considered to be a failure.
+   * Note that task cancellation is not considered to be a failure to this method.
    * If this task has been cancelled then task returned by this method will also
    * be cancelled and recovery function will not be applied.
    *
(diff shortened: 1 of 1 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -626,7 +626,7 @@
    * </pre></blockquote>
    * <img src="doc-files/recover-1.png" height="90" width="462"/>
    * <p>
-   * Note that task cancellation is not considered to be a failure.
+   * Note that task cancellation is not considered to be a failure to this method.
```
karthikbalasub, line 629: I don't think this change is needed. The behavior of cancellation not being considered a failure is not specific to this method.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -847,7 +847,7 @@
 
   /**
    * Creates a new task that will handle failure of this task.
-   * Early completion due to cancellation is not considered to be a failure.
+   * Early completion due to cancellation is not considered to be a failure so it will not be recovered.
```
karthikbalasub, line 850: and the recovery function is not called. (to keep it consistent with the next statement)?

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: In fact cancelled task is failed, right? It contains EarlyFinishException. I think this confuses people. Also in the user guide it compares cancellation to something like "failure" : https://github.com/linkedin/parseq/wiki/User%27s-Guide#cancelling

I personally think we should treat cancellation as special type of failure that .recover, .recoverWith(), .onFailure() don't care about (just like the doc said).

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: Will change.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
mchen07: I see @junchuanwang 's point, it is indeed confusing, especially we have several methods that treats `EarlyFinishException` differently.

Issue #63, opened 2015-10-27
Clarify when parallel tasks are actually executed in parallel


pcmoen:
The documentation should be clarified on what's actually executed in parallel or how to ensure that tasks are executed in parallel.

The documentation have the following statement

> We can use `Task.par()` method to compose tasks to run in parallel

However when running the examples this doesn't seem true for the tasks composed of callables. Using `Task.par(callableTask1, callableTask2)` does not execute these tasks in parallel. This can be easily seen by viewing the Graphviz Timeline of the MergeSort example. None of the parallel operations are executed in parallel.

jodzga:
I agree that documentation does not clarify this and it is probably most misunderstood aspect of parseq. I will provide better description on the wiki.

Thanks for bringing this up!

jodzga:
Jvadoc should at least mention ParSeq execution model.

PR #283, opened 2020-10-01
updating javadoc for Task.par() solving #63

r00pal:
#63 
Updating javadoc for Task.par() when parallel tasks are actually executed in parallel.
Current documentation is kind of misguiding, it should be clearly menntioned that Tasks composed using Task.callable will not be executed in parallel.

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -1434,6 +1434,9 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * will be resolved with results of all tasks as soon as all of them has
    * been completed successfully.
    *
+   * Tasks created using Task.blocking only run in parallel, tasks created using Task.callable are executed
+   * sequentially, although the visualization tools may show the task composition as parallel
+   *
    * <blockquote><pre>
    *  // this task will asynchronously fetch user and company in parallel
    *  // and create signature in a form {@code "<first name> <last name> working for <company>"}


junchuanwang:
> although the visualization tools may show the task composition as parallel

This is actually not true. The video Amen forward(ed?) you should mention that the visualization tool actually tried to explain it is sequential, but the interval is so small that it looks like parallel.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1434,6 +1434,9 @@
    * will be resolved with results of all tasks as soon as all of them has
    * been completed successfully.
    *
+   * Tasks created using Task.blocking only run in parallel, tasks created using Task.callable are executed
+   * sequentially, although the visualization tools may show the task composition as parallel
```
junchuanwang, line 1438: > although the visualization tools may show the task composition as parallel

This is actually not true. The video Amen forward(ed?) you should mention that the visualization tool actually tried to explain it is sequential, but the interval is so small that it looks like parallel.

aliciatang:
Task.par only create another wrapper task, a ParTask, and when any task submitted to the engine it will be allocated a (single) thread. 
So, when you do Task.blocking( task, engine), the task is submitted to engine directly. 
when you do partask = task.par(tasks), engine.run(partask), the partask will take a single thread, and all tasks within it will run on the same thread as the partask. 
I think we should have an option for task.par to run in max allowed number of threads. But that would expose the lower level API and make it harder to understand.

PR #284, opened 2020-10-01
Update ASM to 7.0 for Java 11 compatibility

jessejie7:
Updating ASM for Java 11 compatibility in parseq-lambda-names. The older version of ASM cannot parse Java 11 bytecode so it will fail.


karthikbalasub:
@jessejie7 Close this as we have a better fix implemented?

jessejie7:
done.

PR #291, opened 2021-01-21
Implement scheduleToRun method in Task interface

junchuanwang:
Implement scheduleToRun method in Task interface so the task can be scheduled to current running plan and will be executed asynchronously.

The scheduled task will be run as sibling task.

Example:
```java
    Task.async("MainAsyncTask", () -> {
      final SettablePromise<String> result = Promises.settable();

      System.out.println("I am the MainAsyncTask");

      Task.callable("callable inside", () -> {
        System.out.println("I am a sibling task to MainAsyncTask");
        return "irrelevant";
      }).scheduleToRun();

      result.done("Value");
      return result;
    }).andThen("AndThenAction", 
            Task.action("theActionTask", () -> System.out.println("follow up on Main task")) 
    ).scheduleOrEngineRun(engine);
```
Screen Shot 2021-01-28 at 12 53 32 AM

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -65,6 +65,8 @@ import com.linkedin.parseq.trace.TraceBuilder;
 public interface Task<T> extends Promise<T>, Cancellable {
   static final Logger LOGGER = LoggerFactory.getLogger(Task.class);
 
+  static ThreadLocal<Context> _executionContext = new ThreadLocal<>(); //ThreadLocal.withInitial(null);
+
   static final TaskDescriptor _taskDescriptor = TaskDescriptorFactory.getTaskDescriptor();
 
   //------------------- interface definition -------------------
@@ -122,6 +124,45 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * @param predecessors that lead to the execution of this task
    */
   void contextRun(Context context, Task<?> parent, Collection<Task<?>> predecessors);
+
+  /**
+   * The method will attempt to schedule calling method to a running plan.
+   *
+   * This method should be called inside a lambda function,
+   *   so that when it was called, there are already plans started running:
+   *
+   * <pre>
+   *     Task{@code <Void>} logging(String str) {
+   *         return Task.blocking("Logging", () -> {
+   *          sendToLoggingServer(str);
+   *         });
+   *     }
+   *
+   *     Task{@code <String>} fetchUrl = fetchUrl(url);
+   *     fetchUrl.andThen((result) -> {
+   *      logging(result).scheduleAndRun();
+   *      AnalyzeResult(result);
+   *     });
+   *
+   * </pre>
+   *
+   * This method attempt to schedule the calling task to current plan's scheduler
+   *    and it can be executed asynchronously
+   *
+   * If the method was called when no plans are running,  the calling cannot be scheduled
+   *   to any plans, and it will throws Exception in this case.
+   *
+   * @throws UnsupportedOperationException
+   */
+  default void scheduleAndRun() throws UnsupportedOperationException {
+    Context ctx = _executionContext.get();
+
+    if (ctx == null) {
+      throw new UnsupportedOperationException();
+    }
+
+    ctx.scheduleAndRun(this);
+  }
 
   /**
    * Returns the ShallowTrace for this task. The ShallowTrace will be
@@ -1327,8 +1368,11 @@ public interface Task<T> extends Promise<T>, Cancellable {
   public static <T> Task<T> async(final String name, final Callable<Promise<? extends T>> callable) {
     ArgumentUtil.requireNotNull(callable, "callable");
     return async(name, context -> {
+      _executionContext.set(context);
       try {
-        return callable.call();
+        Promise<? extends T> resultPromise = callable.call();
+        _executionContext.remove();
+        return resultPromise;
       } catch (Throwable e) {
         return Promises.error(e);
       }
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -64,7 +64,7 @@ import com.linkedin.parseq.trace.TraceBuilder;
 public interface Task<T> extends Promise<T>, Cancellable {
   static final Logger LOGGER = LoggerFactory.getLogger(Task.class);
 
-  static ThreadLocal<Context> _executionContext = new ThreadLocal<>(); //ThreadLocal.withInitial(null);
+  static ThreadLocal<Context> _executionContext = new ThreadLocal<>(); // null initialized
 
   static final TaskDescriptor _taskDescriptor = TaskDescriptorFactory.getTaskDescriptor();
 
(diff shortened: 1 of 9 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -123,6 +125,45 @@
...
+   * If the method was called when no plans are running,  the calling cannot be scheduled
+   *   to any plans, and it will throws Exception in this case.
+   *
+   * @throws UnsupportedOperationException
+   */
+  default void scheduleAndRun() throws UnsupportedOperationException {
+    Context ctx = _executionContext.get();
+
+    if (ctx == null) {
+      throw new UnsupportedOperationException();
```
karthikbalasub, line 161: This might be a problem. 
The behavior we need is an API that would use current context if it exists, else execute in a new plan.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -147,21 +161,46 @@
    * </pre>
    *
    * This method attempt to schedule the calling task to current plan's scheduler
-   *    and it can be executed asynchronously
+   *    and it can be executed asynchronously.
    *
    * If the method was called when no plans are running,  the calling cannot be scheduled
    *   to any plans, and it will throws Exception in this case.
    *
+   * Note the task calling {@link#scheduleAndRun} would be running as a sibling task of the
```
mchen07: Forgot to update comment?

junchuanwang: Hi Min I think I have updated this (the version you are checking is an outdated one)

subprojects/parseq/src/main/java/com/linkedin/parseq/internal/ContextImpl.java
```diff
@@ -121,6 +120,40 @@ public void run(final Task<?>... tasks) {
     }
   }
 
+  private void submitToExecutorAsync(Task<?> task) {
+    // Spawn a subContext to drive the running of submitted task later
+    final ContextImpl subContext = new ContextImpl(_planContext, task, _parent, _predecessorTasks);
+    // Note:
+    // (1) We allow task async submission even current task is considered done(promise resolved),
```
mchen07, line 127: will this impact the criteria used in `ParSeqUnitTestHelper.runAndWait` and `runAndWaitForPlanComplete`? In another word, does this current plan need to wait for this task to finish to count the plan complete? since this will impact `Parseq PlanCompletionListener` triggering. 

Also, how is this API different from existing `side-effect` API? My take is that it differs from it in that it is not attached to a particular plan?

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
mchen07: Example code is not updated

subprojects/parseq/src/main/java/com/linkedin/parseq/internal/ContextImpl.java
junchuanwang: The idea is to add a task to execution when the user executes the code in the lambda (at which time a plan is already started).

For side-effect to run, you still need to as the engine to trigger that, am I understanding this correctly?

Yes it will impact PlanCompletionListener (this task need to be done to trigger the planCompletion, so I think this is another difference?

mchen07: No, sideeffect is taken into consideration in `PlanCompletionListener`, only when side-effect task is finished, the listener is triggered for the plan.

Sorry, I still don't understand the motivation here: what do you mean "execution when the user executes the code in the lambda (at which time a plan is already started)."? What scenario cannot be done through current task API? I am not sure that I have a full grasp on the issue. I thought that you want to introduce a way to arbitrarily submit an orphan task to executor to run, but I don't know your expected outcome on how this async task will impact current plan or any plan it will be attached to?

junchuanwang: > "I thought that you want to introduce a way to arbitrarily submit an orphan task to executor to run" 
That is actually right.

> "What scenario cannot be done through current task API?"
Please see the issue description. I have created a task and scheduled to run it directly inside the lambda function. This is one goal. The task can also be created inside the lambda function of a `andThen` method (not shown in the description), so it can reuse parent task result, which is another goal. Similar thing will be complicated to do if to wrap it using `Task.async` and `Context.After().run()`. And as a matter of fact, I don't expect user to do touch `Context` at all (although some users still do). This API is context transparent to the API user.

> "how this async task will impact current plan or any plan it will be attached to"
Actually the parent/child relationship does not  seem to really matter too much in this regard, I only want the created task to be run as part of the plan.

mchen07: Why cannot they achieve the issue described using Task.withSideEffect https://javadoc.io/doc/com.linkedin.parseq/parseq/latest/com/linkedin/parseq/Task.html?  I still didn't get. Lamda or not is just a syntactic sugar on the API, I am still puzzled here.

junchuanwang: Syntax sugar, probably yes, but it will make a difference. I wish we could provide coroutine-alike experience to the end users, but we could not. So this `scheduleToRun` syntax sugar somewhat helps.

The main difference is that the created task returend by Task.withSideEffect will still need to be triggered by `engine.run()`, right? But this `scheduleToRun` is to schedule the task asynchronously and dynamically. To make an analogy, previously it was like run a computational graph by analyzing it statistically and then run (using Engine.run()) and now we are trying to make it dynamic. This api would make things otherwise need to be achieved by `sideEffect` easier. (similar to how the market favors PyTorch over Tensorflow 1.0? )

In order to prevent XY problem, the actual use case is for the universal client:
We are going to expose a Rest.Li implementation of `CompletionStage`, For example the user can put their logic inside the completationStage returned: Comparing the following two method
```java
CompletationStage method1() {
  CompletionStage album1 = Albums.get(1);
  return album1;
}

CompletationStage method2() {
  CompletionStage album1 = Albums.get(1);

  album1.thenCompose((album1) -> {
    log.info(
      AlbumDetail.get(album1.getSerialNumber())
      .toCompletableFuture().get()
    );


  Albums.get(2).thenCompose((albums2) -> {
    Singers.get(album1.getSinger()).thenCompose({
    (singer) -> updateDB(singerId,
      AlbumDetail.get(album2.getSerialNumber().toCompletableFuture.get()));
      })
    });

  });
  return album1;
}
```
When the user run either of the above code(either method1 or method2), the returned completionStage should already started resolving the values.
When the `thenCompose` becaming nested inside multiple layers, the difficulties of writing this using `sideEffect` will be bigger.

mchen07: also, Task.blocking is also serving the similar purpose.

junchuanwang: @mchen07 Hi Min, 
(1) First, how they are invoked are different:
the Task.blocking does serve the same purpose, but again you will have difficulty to trigger it in a lambda function other than manually use `engine.run` to invoke the task.

(2) Second, this API is for all ParSeq tasks. The purpose of blocking task and other parseq tasks are different. For mission critical blocking task, it will be best to run it in another executor service, so wrapping it in a blocking task and run in separate executorservice by calling `engine.run` made sense, but the user of this `scheduleToRun` will be calling it for tasks of different purpose.

Issue #289, opened 2020-11-17
Should deprecate Task.blocking()


junchuanwang:
I recommend we change `Task.blocking()` to `Task.runInExecutor()`
and `Task.Callable()` to `Task.runCallableInCurrentThread()`

BrianPin:
If we change, what about the existing users that will require them to chaneg?
These functions are actually task creation, I think if we can consider:
1. `Task.createTaskInThreadPool`
2. `Task.createSimpleBlockingTask`

how about that

JieyingXu:
> I recommend we change `Task.blocking()` to `Task.runInExecutor()` or `Task.callableInexecutor`

I totally agree with this to be less confusing.

PR #292, opened 2021-01-29
Rename Task.blocking to Task.runInExecutor

junchuanwang:
ParSeq beginner users are often confused about the name of "Task.blocking". It is even more confusing if comparing with other two Task.callable, Task.async, Task.action, Task.value.
(I personally think Task.async confusing as well..)

By analogy, Task.callable means to create a task with a callable. Task.value means to create a task with a value. I think `Task.runInExecutor` is a better user-facing name.  It means to create a task that would "run in executor".

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -1079,7 +1078,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * Creates a new task that have a value of type {@code Void}. Because the
    * returned task returns no value, it is typically used to produce side effects.
    * It is not appropriate for long running or blocking actions. If action is
-   * long running or blocking use {@link #blocking(String, Callable, Executor) blocking} method.
+   * long running or blocking use {@link #runInExecutor(String, Callable, Executor) blocking} method.
    *
    * <blockquote><pre>
    * // this task will print "Hello" on standard output
@@ -1167,7 +1166,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * from the supplied callable. This task is useful when doing basic
    * computation that does not require asynchrony. It is not appropriate for
    * long running or blocking callables. If callable is long running or blocking
-   * use {@link #blocking(String, Callable, Executor) blocking} method.
+   * use {@link #runInExecutor(String, Callable, Executor) blocking} method.
    *
    * <blockquote><pre>
    * // this task will complete with {@code String} representing current time
@@ -1313,7 +1312,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    *
    * This method is not appropriate for long running or blocking callables.
    * If callable is long running or blocking use
-   * {@link #blocking(String, Callable, Executor) blocking} method.
+   * {@link #runInExecutor(String, Callable, Executor) blocking} method.
    * <p>
    *
    * @param <T> the type of the return value for this task
@@ -1403,7 +1402,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
    * @return a new task that will submit the callable to given executor and complete
    * with result returned by that callable
    */
-  public static <T> Task<T> blocking(final String name, final Callable<? extends T> callable, final Executor executor) {
+  public static <T> Task<T> runInExecutor(final String name, final Callable<? extends T> callable, final Executor executor) {
     ArgumentUtil.requireNotNull(callable, "callable");
     ArgumentUtil.requireNotNull(callable, "executor");
     Task<T> blockingTask = async(name, () -> {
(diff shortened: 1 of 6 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1403,7 +1402,7 @@
    * @return a new task that will submit the callable to given executor and complete
    * with result returned by that callable
    */
-  public static <T> Task<T> blocking(final String name, final Callable<? extends T> callable, final Executor executor) {
+  public static <T> Task<T> runInExecutor(final String name, final Callable<? extends T> callable, final Executor executor) {
```
aman1309, line 1405: this is backward incompatible change. why do we need to do this? can we add additional API and deprecate this or just update the java doc to make it less confusing.

aman1309 (changes requested):

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1422,13 +1421,25 @@
...
-   * Equivalent to {@code blocking("blocking", callable, executor)}.
-   * @see #blocking(String, Callable, Executor)
+   * Equivalent to {@code runInExecutor("runInExecutor", callable, executor)}.
+   * @see #runInExecutor(String, Callable, Executor)
    */
+  public static <T> Task<T> runInExecutor(final Callable<? extends T> callable, final Executor executor) {
+    return runInExecutor("runInExecutor: " + _taskDescriptor.getDescription(callable.getClass().getName()), callable, executor);
+  }
+
+  @Deprecated
```
karthikbalasub, line 1434: Comment on the replacement to use.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1422,13 +1421,25 @@
...
   public static <T> Task<T> blocking(final Callable<? extends T> callable, final Executor executor) {
-    return blocking("blocking: " + _taskDescriptor.getDescription(callable.getClass().getName()), callable, executor);
+    return runInExecutor("runInExecutor: " + _taskDescriptor.getDescription(callable.getClass().getName()), callable, executor);
+  }
+
+
+  @Deprecated
```
karthikbalasub, line 1443: Comment on the replacement to use.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: The function with same name still exists and was annotated as deprecated. 

Isn't that what you meant?

aman1309: oops didn't see below. diff showed replacing name. please ignore.

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
aman1309, line 1431: can you add java doc to redirect to new method

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1313,7 +1312,7 @@
    *
    * This method is not appropriate for long running or blocking callables.
    * If callable is long running or blocking use
-   * {@link #blocking(String, Callable, Executor) blocking} method.
+   * {@link #runInExecutor(String, Callable, Executor) blocking} method.
```
aman1309, line 1315: update blocking to runInExecutor. same for other places

subprojects/parseq/src/test/java/com/linkedin/parseq/TestTaskFactoryMethods.java
```diff
@@ -94,7 +94,7 @@ public void testAsyncWithContext() {
   public void testBlocking() {
```
aman1309, line 94: update test name

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: I think here the word `blocking` does not need to be replaced.

aman1309: text will point to #blocking method but link for new method in parsed doc. It will be confusing since we can't remove blocking method

karthikbalasub (commented):
Though the naming was bad before, the need to pass executor as an argument would've given enough hint to the user about the behavior. Also, most of the existing users would now be familiar with Task::blocking behavior. So deprecating it might cause push back from such users. 
What I'm trying to say is, this is one of those cases where you cannot satisfy every user. It comes down to personal preferences.. 

Another ambiguity is that "run" might imply the returned task is running already. However, the task still needs to be run by calling Engine::run. Maybe use "runCallableInExecutor" ?

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: udpated

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
junchuanwang: udpated

junchuanwang:
> Though the naming was bad before, the need to pass executor as an argument would've given enough hint to the user about the behavior. Also, most of the existing users would now be familiar with Task::blocking behavior. So deprecating it might cause push back from such users.
> What I'm trying to say is, this is one of those cases where you cannot satisfy every user. It comes down to personal preferences..
> 
> Another ambiguity is that "run" might imply the returned task is running already. However, the task still needs to be run by calling Engine::run. Maybe use "runCallableInExecutor" ?


@karthikbalasub  I am not worried about existing user familiar with Task::blocking thing, becuase I don't believe there will be anyone who would not understand what happened after seen the java doc. Push back? maybe, but I think the team has the final say and that's why I am checking if the team would agree with me.

For the naming, maybe just "calalbleInExecutor"?

evanw555:
1. In the method body of the new method, I still see a `blockingTask.getShallowTraceBuilder().setTaskType(TaskType.BLOCKING.getName());`, is this something that may cause confusion or inconsistency, since it's no longer `Task.blocking` yet it uses `TaskType.BLOCKING`?
2. On Karthik's point, the fact that there's an executor argument may be enough of a hint. If we're saying it's just the same as callable but with an executor, can we not just overload `Task.callable`? `Task.callable(name, callable, executor)`? May not make sense to pack redundant info into the method name. This would arguably be cleaner.

junchuanwang:
@karthikbalasub @evanw555 I do see benfit of using `Task.callable(name, callable, executor)` (it is cleaner).

But the problem is it seems that in ParSeq history, callable and callableInExecutor (or blocking, as its previous name) were made distinct.

The fact is in history, there was `asyncCallableTask`, than the `Task.blocking`. So we are kind of reverting things back

junchuanwang:
Link to https://github.com/linkedin/parseq/issues/289

Issue #287, opened 2020-10-13
andThen overloading is confusing


junchuanwang:
We have two `andThen` function available. One takes a consumer and second takes a Task.
They have different semantics and behavior.

 https://github.com/linkedin/parseq/wiki/User's-Guide#sequential-composition

A misuse case is that user might use the variant that takes a Consumer function to return the task and want that task to be run, because any object can be cast to void, so the compiler won't issue warning of the consumer function returns the task

Solution can be 
(1) add a new interface, e.g. something like `andThenTask`
(2) add additional logic to see if the return value from the Consumer function is of Task type and execute it (this requires another overloading `andThen` to take a Function<T, Task<R>>

Linked Issue:
https://github.com/linkedin/parseq/issues/207

aliciatang:
+1 andThen use be used to indicate order of execution. 
And the fact that when it takes a Task but does not pass on the result from the previous task is unexpected.

junchuanwang:
https://github.com/linkedin/parseq/pull/291 can be complimentary solution to #287 and #207

BrianPin:
Good on identifying the misuse. +1 for slight preference on the resolution (2)
But I feel there is a common understanding among what `consumer` means, so it can return a task sounds a bit not common and different, so I suggest never expect consumer return anything

junchuanwang:
option (2) is not viable: the Consumer and Function when they apply to the lambda function, they have same erased type. So the lambda function will find two ambiguous overloading.

Also same behavior is achievable using flatMap.

I think we should educate the users instead.

PR #293, opened 2021-01-29
Add another "andThen" method which able to take a function that could return another task

junchuanwang:
Proposed solution to https://github.com/linkedin/parseq/issues/287

Users seem not be confused by use of `map` and `andThen(Task)` in general, but sometimes confused by `andThen(Consumer)`

This is nothing more than a syntax sugar.

I am asking for comments now before adding tests.


junchuanwang:
Decide to close the PR and try to educate users instead.

PR #295, opened 2021-02-16
Add "transformWith" to Task API which performs transformation by executing a task

junchuanwang:
Add "transformWith" to the Task API, that takes a function which returns a task, so that the transformation can be done by executing that task.

A main use case is for making the transformation a task so that can be executed asynchronously


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -845,6 +845,66 @@
     return transform("transform: " + _taskDescriptor.getDescription(func.getClass().getName()), func);
   }
 
+  /**
+   * Create a new task that will transform the result of this task.
+   * Returned task will complete with value calculated by a task returned by the function.
+   *
+   * This is similar to {@link #transform(String, Function1)} except the transformation is done by executing the task returned
+   * by the function.
```
mchen07, line 853: can we have a tracevis flow explanation to this method like the one for existing `transform` API? As well as some code example usage to demonstrate why this new API is useful?

junchuanwang: added

karthikbalasub (approved):

PR #296, opened 2021-03-17
=promise Optimize `toCompletableFuture` implantation if it's already …

hepin1989:
Optimize `toCompletableFuture` implantation if it's already completed.


junchuanwang:
Hi @hepin1989 Thanks for submitting the change!
May I know the motivation of the change? From what I see, there is quite limited improvement.

Since the `addListner` also handles the case of "resolved promise" immediately in its function logic, which is equivalent to the "isDone()" branch you are proposing.

hepin1989:
@junchuanwang Ha, busy at work, The purpose is to reduce a closure creation.

PR #299, opened 2021-05-30
Adding Task.recover(String, Class<X extends Throwable>, func) method

sleepytomcat:
From my experience, it's a repeating pattern to provide different recovery for different types of failures. Currently, it looks like this:
```
Task<Response> task = createResponseTask().recover(ex -> {
                                           if (failure instanceof BadRequestException) {
                                                    return new Response(HTTP_400, ex.getMessage());
                                           }
                                           else if (failure instanceof AuthException) {
                                                    return new Response(HTTP_403, ex.getMessage());
                                           }
                                           else if (failure instanceof ServiceException) {
                                                    return new Response(HTTP_500, ex.getMessage());
                                           });
```

With this introduction of 'recover with failure filter' this same code would get shorter and more concise:

```
Task<Response> task = createResponseTask()
                        .recover(BadRequestException.class, ex -> new Response(HTTP_400, ex.getMessage()))
                        .recover(AuthException.class, ex -> new Response(HTTP_403, ex.getMessage());
                        .recover(ServiceException.class, ex -> new Response(HTTP_500, ex.getMessage());
```

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -637,11 +637,46 @@ public interface Task<T> extends Promise<T>, Cancellable {
    */
   default Task<T> recover(final String desc, final Function1<Throwable, T> func) {
     ArgumentUtil.requireNotNull(func, "function");
+    return recover(desc, Throwable.class, func);
+  }
+
+  /**
+   * Equivalent to {@code recover("recover", func)}.
+   * @see #recover(String, Function1)
+   */
+  default Task<T> recover(final Function1<Throwable, T> func) {
+    return recover("recover: " + _taskDescriptor.getDescription(func.getClass().getName()), Throwable.class, func);
+  }
+
+  /**
+   * Creates a new task that will handle failure of this task.
+   *
+   * Early completion due to cancellation is not considered to be a failure and the recovery function is not called.
+   * If this task competed with failure, and failure type matches the exception class the provided, that failure recovery
+   * function is called.
+   *
+   * If this task completes successfully, then recovery function is not called.
+   *
+   * Note that the first call to <code>recover()</code> with matching exception class provided would result in a recovered Task.
+   * For example, observe the code below:
+   * <pre><code>
+   * Task&lt;String&gt; task = Task.failure(new RuntimeException())
+   *     .recover("try recovering RuntimeException", RuntimeException.class, e -&gt; "recovered") // results in recovery, as exception class matches
+   *     .recover("try recovering Throwable", Throwable.class, e -&gt; "no action"); // no action, as already recovered at previous step
+   * </code></pre>
+   * @param desc description of a recovery function, it will show up in a trace
+   * @param exceptionClass exception class, defines which types of failures would be attempted to recover with {@code func}
+   * @param func recovery function which can complete task with a value depending on failure of this task
+   * @return a new task which can recover from failure of this task
+   */
+  default <X extends Throwable> Task<T> recover(final String desc, final Class<X> exceptionClass, final Function1<? super X, ? extends T> func) {
+    ArgumentUtil.requireNotNull(func, "function");
+    ArgumentUtil.requireNotNull(exceptionClass, "exception class");
     return apply(desc, (src, dst) -> {
       if (src.isFailed()) {
-        if (!(Exceptions.isCancellation(src.getError()))) {
+        if (!(Exceptions.isCancellation(src.getError())) && exceptionClass.isInstance(src.getError())) {
           try {
-            dst.done(func.apply(src.getError()));
+            dst.done(func.apply(exceptionClass.cast(src.getError())));
           } catch (Throwable t) {
             dst.fail(t);
           }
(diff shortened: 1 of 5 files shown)


evanw555:
We see that there may some at least some benefit in this API addition, however we're concerned about the expected behavior here. What happens in this case?
```
Task<Response> task = createResponseTask()
                        .recover(Throwable.class, ex -> new Response(HTTP_400, ex.getMessage()))
                        .recover(FooThrowable.class, ex -> new Response(HTTP_403, ex.getMessage());
                        .recover(FooSubclassThrowable.class, ex -> new Response(HTTP_500, ex.getMessage());
```
Where `FooSubclassThrowable` subclasses `FooThrowable`, and `FooThrowable` subclasses `Throwable`. Will all of them be triggered? And what about in the reverse order? The lack of test coverage for cases like this leaves this ambiguity.

When viewed in this light, the original usage pattern (although slightly more clunky) is a lot more clear in terms of expected behavior, and thus is preferable.

If you're willing to clearly define this behavior and add more thorough test coverage, then we can consider this contribution further. Otherwise, we believe it wouldn't be wise to merge this as-is.

sleepytomcat:
@evanw555 , thank you for review! Let me think more of it, and get back with PR update.
I'll convert this PR to draft.

sleepytomcat:
@evanw555 , I've added additional test clauses to ensure the behavior for exception subclasses is as expected, also updated javadoc to explicitly describe this behavior in documentation.

Overall, the idea of multiple `recover()` matches the behavior of Java's `try`/`catch` pair, and from some perspective may be observed as 'natural'. For example, this is how recovery may be handled in 'plain' Java:
```
try {
     String result = methodThatCanThrowAssoredExceptions();
     return result;
} catch (RuntimeException ex) {
     return "got RuntimeException, recovered";
} catch (Throwable ex) {
     return "got Throwable, recovered";
}
```
For the sake of example, let's say `methodThatCanThrowAssoredExceptions()` has thrown `RuntimeException`. In 'plain' Java that would result in `return "got RuntimeException, recovered";` code executed, but _not_ the clause for `Throwable`, even though `Throwable` is a superclass for `RuntimeException` and is observed as a match.

Similarly, for `Task.recover()` we'd expect the same flow pattern:
```
Task<String> task = Task.failure(new RuntimeException())
     .recover("try recovering RuntimeException", RuntimeException.class, e -> "recovered") // results in recovery, as exception class matches
     .recover("try recovering Throwable", Throwable.class, e -> "no action"); // no action, as already recovered at previous step
```

For comparison, this behavior was implemented in other popular functional programming async libraries, for example:

1. Vavr `Future`, [code here](https://github.com/vavr-io/vavr/blob/master/src/main/java/io/vavr/concurrent/Future.java#L1082)
2. Scala `Future`, [code here](https://www.scala-lang.org/api/2.12.2/scala/concurrent/Future.html) — please note `recover` accepts `PartialFunction` which behaves exactly as an exception filter.

subprojects/parseq/src/test/java/com/linkedin/parseq/AbstractTaskTest.java
```diff
@@ -96,6 +97,34 @@ public void testRecover(int expectedNumberOfTasks) {
     runAndWait("AbstractTaskTest.testRecoverFailure", failure);
     assertEquals((int) failure.get(), -1);
     assertEquals(countTasks(failure.getTrace()), expectedNumberOfTasks);
+
+    Task<Integer> recoveredWithExceptionFilter = getFailureTask()
```
evanw555, line 101: To keep this test case from becoming too bloated, you should split this up into separate test methods.

subprojects/parseq/src/test/java/com/linkedin/parseq/AbstractTaskTest.java
```diff
@@ -96,6 +97,34 @@ public void testRecover(int expectedNumberOfTasks) {
...
+    assertEquals(countTasks(recoveredWithExceptionFilter.getTrace()), expectedNumberOfTasks);
+
+    Task<Integer> notRecoveredWithExceptionFilter = getFailureTask()
+            .map("strlen", String::length)
+            .recover("", ArithmeticException.class, e -> -1);
+    runAndWaitException("AbstractTaskTest.testRecoverWithExceptionFilterFailure", notRecoveredWithExceptionFilter, RuntimeException.class);
+    assertTrue(notRecoveredWithExceptionFilter.isFailed());
+    assertEquals(countTasks(recoveredWithExceptionFilter.getTrace()), expectedNumberOfTasks);
+
+    Task<String> recoveredWithExceptionFilterAndSubclassesForward = getFailureTask() // Failed with RuntimeError
```
evanw555, line 115: This is good, but there are two scenarios that should also be tested:
1. Second recover is selected. e.g. `BarException` is thrown, first recover is `FooException` (unrelated exception), second exception is `BarException`.
2. Case where the first recover itself results in an exception, would it then be caught by the second recover?

sleepytomcat: Fixed, thank you @evanw555 .

subprojects/parseq/src/test/java/com/linkedin/parseq/AbstractTaskTest.java
sleepytomcat: Test cases for 'recovering with exception filter' were extracted into separate `@Test` method.

evanw555:
LGTM, but I request that @junchuanwang also take a look due to his expertise

subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -637,11 +637,46 @@
...
+   * @param func recovery function which can complete task with a value depending on failure of this task
+   * @return a new task which can recover from failure of this task
+   */
+  default <X extends Throwable> Task<T> recover(final String desc, final Class<X> exceptionClass, final Function1<? super X, ? extends T> func) {
+    ArgumentUtil.requireNotNull(func, "function");
+    ArgumentUtil.requireNotNull(exceptionClass, "exception class");
     return apply(desc, (src, dst) -> {
       if (src.isFailed()) {
-        if (!(Exceptions.isCancellation(src.getError()))) {
+        if (exceptionClass.isInstance(src.getError()) && !(Exceptions.isCancellation(src.getError()))) {
```
junchuanwang, line 677: swap the two evaluation so we can leverage short-circuting

PR #306, opened 2021-09-13
Add interop between Guava's ListenableFuture and Parseq Task

karthikrg:


subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Utility methods to convert between Parseq {@link Task} and Guava's {@link ListenableFuture}.
+ */
+public class ListenableFutureUtil {
+
+  private ListenableFutureUtil() {
+    // Prevent instantiation.
+  }
+
+  public static <T> Task<T> fromListenableFuture(ListenableFuture<T> future) {
```
mchen07, line 25: Is it overkill to create a subproject for this? 
Why not follow the similar convention we used to bridge `CompletionStage`.

Promise.toCompletionStage
Task.fromCompletionStage

karthikrg:
Guava is notorious for backward incompatible changes even in minor
versions. I don't want issues coming up in projects that use parseq unless
they really want to use this guava bridging.

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
```diff
@@ -0,0 +1,133 @@
...
+      @Override
+      public boolean setException(Throwable ex) {
+        if (!task.isDone() && ex instanceof CancellationException) {
+          task.cancel((CancellationException) ex);
+        }
+        return super.setException(ex);
+      }
+    };
+
+    // Setup forward event propagation Task -> ListenableFuture.
```
junchuanwang, line 76: nit: is the wording correct? or should it be backward (corresponding to the "forward" in "fromListenableFuture"

subprojects/parseq-guava-interop/src/test/java/com/linkedin/parseq/guava/ListenableFutureUtilTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.testng.annotations.Test;
+
+
+/**
+ * Unit tests for {@link ListenableFutureUtil}
+ */
+public class ListenableFutureUtilTest {
+
+  @Test
+  public void testFromListenableFuture() {
```
junchuanwang, line 17: do you think we can easily add similar test case with an actual delayed setup? Something similar to what we do
https://github.com/linkedin/rest.li/blob/master/restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java#L155

https://github.com/linkedin/rest.li/blob/master/restli-client/src/test/java/com/linkedin/restli/client/TestParSeqBasedCompletionStage.java#L181

junchuanwang (approved):
It will be great if we can enhance the tests.

Also thanks for providing this solution using nested task class

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
woodyiorl: I also prefer it in separate subproject because Guava might not be so commonly used.

junchuanwang: I also prefer it in separate subproject because Guava might not be so commonly used.

karthikrg: Guava is notorious for backward incompatible changes even in minor
versions. I don't want issues coming up in projects that use parseq unless
they really want to use this guava bridging.

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
karthikrg: I just find this clearer to read... and it is forward and backward right?

subprojects/parseq-guava-interop/src/test/java/com/linkedin/parseq/guava/ListenableFutureUtilTest.java
karthikrg: I don't think that is going to achieve anything. We will basically be testing that the behavior of listenablefuture and parseq task is consistent in a multi-threaded environment... which has already been tested in numerous places... so I am assuming it as an idiom :)

PR #309, opened 2021-11-23
Add rate limit retry policy

misoukrane:
I added a RateLimit retry policy to help with controlling the qps of retries.

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java
@@ -1,6 +1,8 @@
 package com.linkedin.parseq.retry;
 
 import com.linkedin.parseq.retry.backoff.BackoffPolicy;
+import com.linkedin.parseq.retry.termination.RateLimiter;
+import com.linkedin.parseq.retry.termination.RequireAny;
 import com.linkedin.parseq.retry.termination.RequireEither;
 import com.linkedin.parseq.retry.termination.TerminationPolicy;
 
@@ -70,6 +72,29 @@ public interface RetryPolicy {
     TerminationPolicy terminationPolicy = new RequireEither(TerminationPolicy.limitAttempts(attempts), TerminationPolicy.limitDuration(duration));
     return new RetryPolicyBuilder().
         setName("RetryPolicy.LimitAttemptsAndDuration").
+        setTerminationPolicy(terminationPolicy).
+        setBackoffPolicy(BackoffPolicy.constant(backoff)).
+        build();
+  }
+
+  /**
+   * Retry policy with configurable number of retries, limited total duration, and limited rate of the encompassing task.
+   *
+   * @param attempts Total number of attempts (the number of retries will be that minus 1).
+   * @param duration Total duration of the task. This includes both the original request and all potential retries.
+   * @param rateLimiter The limiter to control the rate of retries per second.
+   * @param backoff The constant delay (in milliseconds) between retry attempts.
+   */
+  static RetryPolicy attemptsAndDurationAndRate(int attempts, long duration, RateLimiter rateLimiter, long backoff) {
+
+    TerminationPolicy terminationPolicy = new RequireAny(
+      TerminationPolicy.limitAttempts(attempts),
+      TerminationPolicy.limitDuration(duration),
+      TerminationPolicy.limitRate(rateLimiter)
+    );
+
+    return new RetryPolicyBuilder().
+        setName("RetryPolicy.LimitAttemptsAndDurationAndRate").
         setTerminationPolicy(terminationPolicy).
         setBackoffPolicy(BackoffPolicy.constant(backoff)).
         build();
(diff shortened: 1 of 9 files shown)


subprojects/parseq/src/test/java/com/linkedin/parseq/retry/TestRetryPolicy.java
```diff
@@ -35,6 +37,20 @@ public void testSimpleRetryPolicy()
     assertEquals(task.getError().getMessage(), "current attempt: 2");
   }
 
+  @Test
+  public void testRateLimitedRetryPolicy() throws InterruptedException
+  {
+    RateLimiter rateLimiter = new GuavaRateLimiter(4.0);
+    // wait for the rate limiter to be ready
```
junchuanwang, line 44: why is this needed?

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/LimitRate.java
```diff
@@ -0,0 +1,27 @@
+package com.linkedin.parseq.retry.termination;
+
+/**
+ * A termination policy that limits the rate of retries.
```
junchuanwang, line 4: IMO we should add an brief introduction on how this rate limiter works (sliding window, leaky bucket, etc) because other two retry policy are quite self-explanary (duration, attempts)

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/RetryPolicy.java
```diff
@@ -74,4 +76,27 @@ static RetryPolicy attemptsAndDuration(int attempts, long duration, long backoff
...
+
+  /**
+   * Retry policy with configurable number of retries, limited total duration, and limited rate of the encompassing task.
+   *
+   * @param attempts Total number of attempts (the number of retries will be that minus 1).
+   * @param duration Total duration of the task. This includes both the original request and all potential retries.
+   * @param rateLimiter The limiter to control the rate of retries per second.
+   * @param backoff The constant delay (in milliseconds) between retry attempts.
+   */
+  static RetryPolicy attemptsAndDurationAndRate(int attempts, long duration, RateLimiter rateLimiter, long backoff) {
```
junchuanwang, line 88: Now you added a policy based on three criteria, but I am wondering: does "using rateLimiter" as the only criteria work?

misoukrane: using a ratelimiter by itself would work, but it will be suboptimal for many scenarios. Also, the combination of all these policies make the retry process fairer for all types of requests: retries and original requests.

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/LimitRate.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @author Mohamed Isoukrane (isoukrane.mohamed@gmail.com)
+ */
+public class LimitRate implements TerminationPolicy {
```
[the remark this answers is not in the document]
misoukrane: I was following the same naming convention that was used for duration and attempts there are two classes named: `LimitAttempts` & `LimitDuration`, that's why I went with `LimitRate`.
I think being consistent with existing nomenclature is encourage; if this is not true, let me know what your suggestions is for this class name.

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/RequireAny.java
```diff
@@ -0,0 +1,33 @@
+package com.linkedin.parseq.retry.termination;
+
+/**
+ * A termination policy that signals for termination after any of the specified policies terminate.
+ *
+ * @author Mohamed Isoukrane (isoukrane.mohamed@gmail.com)
+ */
+public class RequireAny implements TerminationPolicy {
```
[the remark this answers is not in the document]
misoukrane: I basically tried to be consistent with the exiting naming convention: `RequireBoth` and `RequireEither`.

subprojects/parseq/src/test/java/com/linkedin/parseq/retry/TestRetryPolicy.java
misoukrane: Guava rate limiter has a warmup period, and it tries to make the permits "distributed smoothly with the delay between individual permits being adjusted to ensure that the configured rate is maintained"
so that sleep is just for the rate limit to warmup.
The guava implementation has more details https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/RateLimiter.html#create(double) .

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/RequireAny.java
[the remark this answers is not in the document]
junchuanwang: thanks. Make sense

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/LimitRate.java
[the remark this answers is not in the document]
junchuanwang: thanks. Make sense

subprojects/parseq/src/main/java/com/linkedin/parseq/retry/termination/LimitRate.java
misoukrane: Instead of adding it to the interface, I'll add the description to the implementation.

junchuanwang (approved):

PR #324, opened 2022-06-02
Add more logs in ListenableFutureUtil

junchuanwang:
Two changes
(1) Add more logs in ListenableFutureUtil to understand the cancel sequence.
(2) when we call BaseTask.cancel(), we need to cancel "future.cancel" no matter what, therefore this statement better not to be short-circuited.


cgtz (approved):
lgtm

PR #327, opened 2022-12-21
Fix cancel() implementation in ListanbleFutureUtil

junchuanwang:
Previous implementation has two problems that fixed in this PR.

1. We now use best effort to cancel GRPC future and that shouldn't impact the return value of `cancelled()` that supposed to indicate base Task cancellation 
2. We should  only cancel GRPC future when base task state transition confirmed to happen.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
           fetch-depth: 2
       - uses: actions/setup-python@v2
         with:
-          python-version: '2.7.5'
+          python-version: '3.9.12'
           cache: 'pip'
       - uses: actions/setup-java@v2
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
           fetch-depth: 2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9.12'
+          python-version: '2.7.18'
       - uses: actions/setup-java@v2
         with:
           distribution: zulu
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,10 @@ jobs:
         with:
           # Need to fetch 2 commits for the PR (base commit and head merge commit) so we can compute the diff
           fetch-depth: 2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '2.7.5'
+          cache: 'pip'
       - uses: actions/setup-java@v2
         with:
           distribution: zulu
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,6 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.9.12'
-          cache: 'pip'
       - uses: actions/setup-java@v2
         with:
           distribution: zulu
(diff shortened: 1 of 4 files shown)


subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
```diff
@@ -41,19 +41,16 @@ public static <T> Task<T> fromListenableFuture(ListenableFuture<T> future) {
...
-              boolean taskCancelResult = super.cancel(rootReason);
-              LOGGER.warn("Future is cancelled, therefore cancelling task, result: {}", taskCancelResult);
-              return taskCancelResult;
+            // <BaseTask>.cancel()'s result indicates whether cancel() successfully trigger state transition to "CANCELLED"
+            // And we should only cancel GRPC future when the transition was conducted.
+            boolean baseStateTransitedByCallingCancel = super.cancel(rootReason);
+            if (baseStateTransitedByCallingCancel && !future.isCancelled()) {
+              boolean futureCancelResult = future.cancel(true);
+              if (!futureCancelResult) {
+                LOGGER.error("Unexpected error: GRPC future was not cancelled but new attempt to cancel also failed.");
```
mchen07, line 50: I don't think that we should error for this. We tried to cancel gRPC future, but if it failed, we should not error here, at most log as warning.

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
```diff
@@ -41,19 +41,16 @@ public static <T> Task<T> fromListenableFuture(ListenableFuture<T> future) {
         new BaseTask<T>("fromListenableFuture: " + Task._taskDescriptor.getDescription(future.getClass().getName())) {
           @Override
           public boolean cancel(Exception rootReason) {
-            if (future.isCancelled()) {
-              boolean taskCancelResult = super.cancel(rootReason);
-              LOGGER.warn("Future is cancelled, therefore cancelling task, result: {}", taskCancelResult);
-              return taskCancelResult;
+            // <BaseTask>.cancel()'s result indicates whether cancel() successfully trigger state transition to "CANCELLED"
+            // And we should only cancel GRPC future when the transition was conducted.
+            boolean baseStateTransitedByCallingCancel = super.cancel(rootReason);
```
mchen07, line 46: Maybe just name it `shouldCancelTask`?

.github/workflows/build.yml
```diff
@@ -18,6 +18,10 @@ jobs:
         with:
           # Need to fetch 2 commits for the PR (base commit and head merge commit) so we can compute the diff
           fetch-depth: 2
+      - uses: actions/setup-python@v2
```
mchen07, line 21: what is this new file? I am not aware of this.

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
rickzx: +1. Easier to understand

subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
rickzx: +1 for warning. Cancellation is best effort, and if the future is already complete, it is likely that it cannot be cancelled

rickzx (changes requested):

.github/workflows/build.yml
junchuanwang: the github action upgraded their python version and is conflicting with existing code in `parseq-tracevis`

rickzx (approved):

mchen07 (approved):

PR #328, opened 2022-12-21
Pin python version in the publish github action

junchuanwang:
Have to pin the python version since new version doesn't work with node-gyp


rickzx (approved):
Thanks!

PR #331, opened 2023-02-28
Upgrade bytebuddy and asm version for JDK 17 and JDK 21 support

JoeJoe1989:
Upgrade bytebuddy and asm version for JDK 17 and JDK 21 support.
Created snapshot and tested against apps with JDK 8, JDK 11, JDK 17 and JDK 21 runtime.


aman1309 (approved):

junchuanwang:
tiny question: should JDK 14 be part of testing plan

JoeJoe1989:
> tiny question: should JDK 14 be part of testing plan

I don't think it is needed because it is not a LTS and we are already testing a higher LTS version (JDK 17), which should encompass the features/changes introduced since JDK 14.

junchuanwang (approved):

PR #336, opened 2023-10-05
Avoid blocking in ASMBasedTaskDescriptor

cbrentharris:
Currently the ASMBasedTaskDescriptor uses ByteBuddy to hook into the class loader to analyze bytecode so that parseq can generate descriptions of tasks that were created via lambda functions to aid in debugging and tracing.

In production, blocking on the lambda can cause a global lockout of threads. This can happen because of the common fork join pool being starved or because the classes being analyzed take a while.

If a class has already been analyzed, we do not need to await the latch because we have a threadsafe collection to access. Classes can be analyzed multiple times in a JVM lifecycle because of class unloading, but we don't care about subsequent analyses, (the bytecode should not change with load -> unload -> load).


jennybkim (approved):

junchuanwang (commented):
I am confused. In the scenario you described, where ` load -> unload -> load `could happen (GC or something). Wouldn't returning from cached result potentially, render incorrect result?

junchuanwang (commented):
>  blocking on the lambda can cause a global lockout of threads. This can happen because of the common fork join pool being starved or because the classes being analyzed take a while.

I am afraid I also have difficulty understanding this. You are trying to say the common fork joining pool starved because of being occupied, but this is different than "global lockout of threads", unless it is the CPU resource itself is really exhausted. But I cannot help wondering what other tasks need to be done before classes are loaded?

Besides my question, I also have doubt what difference your change is making to this situation. In what case does getLambdaClassDescription() have to return before code analysis are done?

cbrentharris:
> I am confused. In the scenario you described, where ` load -> unload -> load `could happen (GC or something). Wouldn't returning from cached result potentially, render incorrect result?
> 
> 
> 
> 
> 
> 

The byte code for a class should never be actually changed during a single JVM lifecycle instance otherwise very odd things would happen

cbrentharris:
> >  blocking on the lambda can cause a global lockout of threads. This can happen because of the common fork join pool being starved or because the classes being analyzed take a while.
> 
> 
> 
> I am afraid I also have difficulty understanding this. You are trying to say the common fork joining pool starved because of being occupied, but this is different than "global lockout of threads", unless it is the CPU resource itself is really exhausted. But I cannot help wondering what other tasks need to be done before classes are loaded?
> 
> 
> 
> Besides my question, I also have doubt what difference your change is making to this situation. In what case does getLambdaClassDescription() have to return before code analysis are done?

So any parseq engine executing a plan that has a lamda will eventually call the ASMTaskedBased descriptor. And in my experience that is nearly all plans.

Parseq generally runs with a very low thread count

Now there are two potential reasons all Parseq threads could be locked

1) when trying to get a descriptor and another analysis is happening, the fork join pool is being used by some third party library and Parseq has to wait and so do all classes. This is worsened if ClassUnloading is enabled for the JVM
2) when trying to get a descriptor and another analysis is happening, the analysis is taking a looong time. This is worsened if ClassHnloading is enabled for the JVM

This change set will not lock out a thread if an analysis has already been done, and will not cause (very long) random complete stoppages in an application (currently seeing this)

subprojects/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
```diff
@@ -71,112 +72,109 @@ static void onExit(@Advice.Argument(0) Class<?> hostClass, @Advice.Argument(1) b
...
+      return Optional.empty();
+    }
+    String name = className.substring(0, slashIndex);
+    String description = _names.get(name);
+
+    // If we have already analyzed the class, we don't need to await
+    // analysis on other lambdas.
+    if (description != null) {
+      return Optional.of(description);
+    }
```
zackthehuman, line 160: There is a slight change in behavior (hard to say how significant it is, though): the previous code treated empty string the same as `null` and returns `Optional.empty()`. This code only does it for null, so it can return "" as a valid name (hypothetically). Can we also map "" to `Optional.empty()` so at least we'll get the class name as a fallback?

zackthehuman (changes requested):

subprojects/parseq-lambda-names/src/main/java/com/linkedin/parseq/lambda/ASMBasedTaskDescriptor.java
cbrentharris: sure thing!

zackthehuman (approved):
Looks good to me. Thanks for the patch and for replacing those pesky tabs with spaces. This is the way.

PR #338, opened 2023-10-17
Cbrentharris/undo refaactor of anonymous class to lambda

cbrentharris:
Create defined class to perform analysis within thread
When refactoring from an Anonymous class to a lambda, an infinite
recursion occured because each analysis was required to then
fork and create a new analysis when it was done because it defined
a new anonymous class.

This PR simply defines the class specifically to avoid it analyzing
itself. A follow up should be to filter anonymous classes unrelated
to parseq

Testing done -- brought it into local repository that was seeing stackoverflow
and no longer see it.


zackthehuman (approved):
Thanks for catching this one.

PR #341, opened 2023-10-25
run sendRequest in an executor

sji01:
This PR adds the option to run sendRequest in a dedicated threadpool.

Although sendRequest is asynchronous, we found that it may still block the scheduling thread for extended amount of time in certain scenarios.

Here is one example we had in production:

image

With this change, we no longer have such blocking behavior in the scheduling thread.

### DIFF
--- a/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/DirectExecutor.java
+++ b/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/DirectExecutor.java
@@ -1,0 +1,21 @@
+package com.linkedin.restli.client;
+
+import java.util.concurrent.Executor;
+
+
+/**
+ * A simple executor implementation that executes the task immediately on the calling thread
+ */
+class DirectExecutor implements Executor {
+
+  private static final DirectExecutor INSTANCE = new DirectExecutor();
+
+  static DirectExecutor getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public void execute(Runnable command) {
+    command.run();
+  }
+}
--- a/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
+++ b/subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.restli.client;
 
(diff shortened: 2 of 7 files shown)


subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/DirectExecutor.java
```diff
@@ -0,0 +1,21 @@
+package com.linkedin.restli.client;
+
+import java.util.concurrent.Executor;
+
+
+/**
+ * A simple executor implementation that executes the task immediately on the calling thread
+ */
+class DirectExecutor implements Executor {
```
junchuanwang, line 9: is this similar to MoreDirectors.directDirector() ?

sji01: Yes. but @zackthehuman prefers not introducing dependencies on Guava.

zackthehuman: Correct; I don't see any other usage of Guava in the `parseq-restli-client` module, and I think it's a bit heavy-handed to introduce it just so we can use the `DirectExecutor`. I see two other options here:

1) Implement a simple `DirectExecutor` (which is what you've done)
2) Make the `_executor` nullable and then if it's `null` just don't use an executor to run `sendRequest` (which is what the code did prior to this PR)

subprojects/parseq-restli-client/src/main/java/com/linkedin/restli/client/ParSeqRestClient.java
```diff
@@ -79,6 +94,7 @@ public class ParSeqRestClient extends BatchingStrategy<RequestGroup, RestRequest
     _requestConfigProvider = requestConfigProvider;
     _requestContextProvider = requestContextProvider;
     _d2RequestTimeoutEnabled = d2RequestTimeoutEnabled;
+    _executor = DirectExecutor.getInstance();
```
aman1309, line 97: can we call the new constructor instead of duplicate assignments here

sji01: Sure. Updated the PR.

aman1309 (approved):
Looks good, thank you for the contribution.

PR #345, opened 2024-01-22
Upgrade async http client

cbrentharris:
The project has moved to a new repo; please refer to https://github.com/apache/pinot/pull/7968/ and https://github.com/AsyncHttpClient/async-http-client/issues/934 for motivation

### DIFF
--- a/subprojects/parseq-tracevis-server/src/main/java/com/linkedin/parseq/JhatHandler.java
+++ b/subprojects/parseq-tracevis-server/src/main/java/com/linkedin/parseq/JhatHandler.java
@@ -18,6 +18,7 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.asynchttpclient.Response;
 import org.codehaus.jackson.JsonNode;
 import org.codehaus.jackson.map.ObjectMapper;
 import org.eclipse.jetty.server.Request;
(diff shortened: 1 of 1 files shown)


aliciatang (approved):
:shipit:

subprojects/parseq-http-client/src/main/java/com/linkedin/parseq/httpclient/HttpClient.java
```diff
@@ -15,9 +19,9 @@ public class HttpClient {
...
    * @return raw http client
    */
-  public static synchronized AsyncHttpClient getNingClient() {
+  public static synchronized AsyncHttpClient getAsyncHttpClient() {
     if (_client.get() == null) {
-      initialize(new AsyncHttpClientConfig.Builder().build());
+      initialize(new Builder().build());
     }
     return _client.get();
   }
```
zackthehuman, line 27: I think we should instead introduce `getAsyncHttpClient()` alongside `getNingClient()` and mark `getNingClient()` as `@Deprecated`. I know the type has changed, and there's nothing we can do about that, but there are users of this code that will break if `getNingClient()` disappears on a minor version bump.

zackthehuman: I need to confer with the team about the other breaking changes.

zackthehuman (changes requested):

subprojects/parseq-http-client/src/main/java/com/linkedin/parseq/httpclient/HttpClient.java
cbrentharris: The only concern I would have with that is if ning needs a different version of netty (or other) that the new project needs. What are your thoughts on that behavior? Would you like to see the dependency resolution for netty?

cbrentharris: [netty-dependencies.txt](https://github.com/linkedin/parseq/files/14040076/netty-dependencies.txt)

PR #348, opened 2024-03-21
Add better future support

zsmoore:
Add support for native completable future input + returns + remove the idea of wrapping in a callable

Right now completionStage is often used with a simple callable () -> future.
Get rid of the callable and just take in the future direct

Also expose CompletableFuture concretely instead of through completions stage

Added tests

### DIFF
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -17,7 +17,6 @@
 package com.linkedin.parseq;
 
 import java.util.Collection;
-
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
--- a/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
+++ b/subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
@@ -1065,7 +1065,7 @@ public interface Task<T> extends Promise<T>, Cancellable {
   default Task<T> withTimeout(final String desc, final long time, final TimeUnit unit) {
     final Task<T> that = this;
     final String taskName = "withTimeout " + time + TimeUnitHelper.toString(unit) +
-            (desc != null ? " " + desc : "");
+        (desc != null ? " " + desc : "");
     final String timeoutExceptionMessage = "task: '" + getName() + "' " + taskName;
     Task<T> withTimeout = async(taskName, ctx -> {
       final AtomicBoolean committed = new AtomicBoolean();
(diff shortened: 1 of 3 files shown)


subprojects/parseq/src/main/java/com/linkedin/parseq/Task.java
```diff
@@ -1355,6 +1356,41 @@ public static <T> Task<T> fromCompletionStage(final Callable<CompletionStage<? e
...
+   *
+   * Returned task will fail if future passed in as a parameter throws
+   * an exception.
+   *
+   * @param <T> the type of the return value for this task
+   * @param desc description of the task, it will show up in a trace
+   * @param future the future to map from when this task is run
+   * @return the new task that will listen to the future and return the data from the underlying future when complete
+   */
+  public static <T> Task<T> fromFuture(final String desc, final CompletableFuture<? extends T> future) {
```
mchen07, line 1371: is existing `fromCompletionStage` not working?

zsmoore: It works, the only annoyance is the callable seems unnecessary.  If we want to keep impl to the interface I can just add a new alternative which accepts a CompletionStage instead of a Callable<CompletionStage>

zackthehuman: Why does the callable seem unnecessary? It's needed to make the behavior of `Task` consistent.

`Task` is lazy. ParSeq needs a way to start a `Task` and thus start the work "inside" the task. One way it can achieve this is through a callable, which returns the thing that needs to run inside the `Task`.

`Future` is not lazy, which means that it doesn't play well with `Task` by default. By deferring the creation or execution of a `Future` (or `CompletionStage`) we can achieve the needed laziness.

junchuanwang: +1 to what Zack said.

But I also don't mind add this additional API if we made the documentations clear. What do you guys think?

zackthehuman: I don't see a compelling use case for this. Providing an API for already-running futures is going to make tracing useless for those tasks, which is only going to cause more confusion (e.g. "why does ParSeq say this Task took 1ms when the Future took 15ms?"). If anything, I'd love to reduce the amount of API ParSeq has, not increase it :)

From what I see here, the problem is not that current API is limiting, it's that it isn't clear why it uses `Callable` to preserve its laziness contract. If this was well understood, I don't think this PR would even exist.

zsmoore: I would agree, it is not clear to me why Callable is needed, we can close this one out then. Thanks for the explanation crew!

mchen07: +1 to what @zackthehuman explained above. I think that @zsmoore got inspired by our existing interop btw Task and ListenableFuture (https://github.com/linkedin/parseq/pull/306), but that indeed raised some questions when people saw that their gRPC calls have started (not lazy).

PR #351, opened 2024-06-11
Fix bridging for get

karthikrg:
We use the delegated promise of the BaseTask instead of creating another one to make sure that get also gets bridged correctly.


subprojects/parseq-guava-interop/src/test/java/com/linkedin/parseq/guava/ListenableFutureUtilTest.java
```diff
@@ -62,6 +62,13 @@ public void testFromListenableFuture() throws Exception {
     Assert.assertTrue(task.isDone());
     Assert.assertTrue(task.isFailed());
     Assert.assertEquals(task.getError().getCause().getClass(), CancellationException.class);
+
+    listenableFuture = new ListenableFutureUtil.SettableFuture<>();
```
mchen07, line 66: I don't understand this change, did line 39-43 already test this path? Only difference is that there is `runUntilComplete(task)`

PR #352, opened 2025-06-16
Update ListenableFutureUtil.java with log message

ayush571995:
This is a bug fix .
The error message seems a little dubious 
The function  is accepting task of Type Future but logs message with the task of type GRPC


subprojects/parseq-guava-interop/src/main/java/com/linkedin/parseq/guava/ListenableFutureUtil.java
```diff
@@ -47,7 +47,7 @@ public boolean cancel(Exception rootReason) {
             if (shouldCancelTask && !future.isCancelled()) {
               boolean futureCancelResult = future.cancel(true);
               if (!futureCancelResult) {
-                LOGGER.warn("Unexpected: GRPC future was not cancelled but new attempt to cancel also failed.");
+                LOGGER.warn("Unexpected: Task {} future was not cancelled but new attempt to cancel also failed.", future.getClass.getName());
```
mchen07, line 50: I am not clear about this change. `ListenableFuture` is from gRPC client stub invocation, that is why previous warning mentioned `GRPC future`, it is actually not a ParSeq task? What extra value `future.getClass.getName()` will be provided here? Most cases it will be `ListenableFuture`?

ayush571995: So my understanding truly from the naming convention of the function is 

1. The class is ListenableFutureUtil is providing utility methods and the implementations or contract are quite generic . the parameters is of Type ListenableFuture hence the errors or warnings should be quite generic .
Giving warning unexpected Grpc future was not cancelled is more concrete 
Rather if the fromListenableFuture function had parameters GrpcFuture this message would have made sense to me 

2. By removing GRPC I just wanted to give more generic message fetching from the task details
I guess you're right , future.getClass.getName() might not be useful here, hence I replaced with task.getName();

mchen07: To me,  we can only make that log message more accurate by 
```
LOGGER.warn("Unexpected: Listenable future was not cancelled but new attempt to cancel also failed.
```
Not sure if you can provide any other further info.

ayush571995: updated

mchen07 (approved):

mchen07:
pending this pR https://github.com/linkedin/parseq/pull/355 to merge to fix build issue.

mchen07 (approved):

zhihaodev:
Could you rebase with master and see if the test can pass?

ayush571995:
Rebased with master

PR #354, opened 2025-07-01
Add BaseEngineParSeqTest.java to make test suite run in parallel with no race conditions

Balaji4:
** Testing done **
Tested the code with `./gradlew clean build`. 

```
BUILD SUCCESSFUL in 12m 6s
113 actionable tasks: 102 executed, 11 up-to-date
```

Also tested with a giant private repo with thousands of unit tests (this base class also improved the test suite execution time by letting us run tests in parallel).

------------------------------------------------

**More details:**
There are two changes in this PR:

**Part 1:**
We currently have `BaseEngineTest` (which runs `setup()` and `teardown()` at method level) and `BaseEngineParTest` (which runs `setup()` and `teardown()` at class level). Both are prone to race condition issues for `ParseqUnitTestHelper` instance as it will be setup and teardown in different tests simultaneously when tests are executed in parallel. 
Added `BaseEngineParSeqTest.java`, which runs `setup()` and `tearDown()` at suite level and will also make sure that the shared `ParseqUnitTestHelper` instance will not have any race condition issues. This will fix all the race condition issues and tests can be executed in parallel at suite level thereby reducing the test suite execution time.
 

**Part 2:** 
`jcenter` is sunset as of Aug 15 2024 (source: https://jfrog.com/blog/jcenter-sunset/). Also, we are using an older version of `build-info-extractor-gradle` (the version doesn't exist in maven and gradle repository). Removing jcenter and updating `build-info-extractor-gradle` (bare minimum changes to fix build issues).

### DIFF
--- a/subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
+++ b/subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2012 LinkedIn, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.linkedin.parseq;
+
+import com.linkedin.parseq.Engine;
+import com.linkedin.parseq.ParSeqUnitTestHelper;
+import com.linkedin.parseq.Task;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import org.testng.annotations.AfterSuite;
+import org.testng.annotations.BeforeSuite;
+
+/**
+ * This is a custom implementation of {@link com.linkedin.parseq.BaseEngineTest}. This custom
+ * implementation is necessary to avoid race-conditions that we have in BaseEngineTest when running tests
+ * in parallel. It's a variant of {@link com.linkedin.parseq.BaseEngineTest}, and has the following
+ * differences:
+ *
+ * <p>1) Setup and tearDown are executed at test-suite level. Whereas {@link
+ * com.linkedin.parseq.BaseEngineTest} executes setup()/teardown() at class level (which can cause
+ * race-conditions if two or more tests from the same class are executed in parallel).
+ *
+ * <p>2) It uses mutex locking mechanism to set up and tear down {@link ParSeqUnitTestHelper} to
+ * avoid race-condition on ParSeqUnitTestHelper instance (it's a common resource among tests).
+ *
+ * <p>3) Default timeout for runAndWait() method is 10 seconds (which is needed by majority of the tests).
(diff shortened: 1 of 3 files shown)


subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright 2012 LinkedIn, Inc
```
karthikrg: nit: 2025

subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final Object MUTEX = new Object();
+  private static final int RUN_AND_WAIT_TIMEOUT_SECONDS = 10;
+  private static final ParSeqUnitTestHelper PAR_SEQ_UNIT_TEST_HELPER = new ParSeqUnitTestHelper();
+
+  /*
+   * Get ParSeqUnitTestHelper
+   *
+   * @return ParSeqUnitTestHelper instance.
+   */
+  protected static ParSeqUnitTestHelper getParSeqUnitTestHelper() {
```
karthikrg, line 54: Why expose this weirdly like this? Just make the instance protected?

subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
```diff
@@ -0,0 +1,155 @@
...
+ * This is a custom implementation of {@link com.linkedin.parseq.BaseEngineTest}. This custom
+ * implementation is necessary to avoid race-conditions that we have in BaseEngineTest when running tests
+ * in parallel. It's a variant of {@link com.linkedin.parseq.BaseEngineTest}, and has the following
+ * differences:
+ *
+ * <p>1) Setup and tearDown are executed at test-suite level. Whereas {@link
+ * com.linkedin.parseq.BaseEngineTest} executes setup()/teardown() at class level (which can cause
+ * race-conditions if two or more tests from the same class are executed in parallel).
+ *
+ * <p>2) It uses mutex locking mechanism to set up and tear down {@link ParSeqUnitTestHelper} to
```
karthikrg, line 37: Why is this static and not scoped per test class?

subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
Balaji4: Updated

subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
Balaji4: I just recalled why I made it static:

I am making it singleton with this logic because I need one instance for the whole test suite so that I can setup and teardown `PAR_SEQ_UNIT_TEST_HELPER` in beforeSuite and afterSuite methods respectively.
BeforeSuite and AfterSuite methods run before any test's constuctor (or its base class's constructor). So making `PAR_SEQ_UNIT_TEST_HELPER` non-static will cause NPE issues.
Regarding why BeforeSuite and AfterSuite methods, I already commented the reason from line 33 to 35.

subprojects/parseq-test-api/src/main/java/com/linkedin/parseq/BaseEngineParSeqTest.java
Balaji4: Same as above, need a singleton instance of ParSeqUnitTestHelper

